PR #5, opened 2024-08-28
chore: celestia migration

irnb:
- Update to latest version of celestia library 
- Adopt new approach of wiring in node framework


steph-mipt (approved):
let's apply `zk fmt`
and LGTM!

PR #7, opened 2024-08-30
feat: `btc_watch` basic structure

steph-mipt:
This PR provides basic initialization of `btc_watch` crate, configs, structs and layer implementation

PR #10, opened 2024-09-03
feat: `btc_watch` basic structure

steph-mipt:
This PR provides basic initialization of btc_watch crate, configs, structs and layer implementation
(reopened #7)


steph-mipt:
I have withheld some changes from the review (#7) until next implementation stages

core/lib/config/src/configs/via_btc_watch.rs
```diff
@@ -0,0 +1,91 @@
...
+    pub rpc_password: String,
+
+    /// Network of the Bitcoin node.
+    pub network: String,
+
+    /// List of transaction IDs to bootstrap the indexer.
+    pub bootstrap_txids: Vec<String>,
+
+    /// Role of the actor. SEQUENCER or VERIFIER.
+    pub actor_role: String,
```
mpavlovic-txfusion, line 31: Probably better to make this an enum

core/node/via_btc_watch/src/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+use crate::message_processors::MessageProcessorError;
+
+mod message_processors;
+
+// re-exporting here isn't necessary, but it's a good practice to keep all the public types in one place
+pub use via_btc_client::types::BitcoinNetwork;
+
+#[derive(Debug)]
+struct BtcWatchState {
+    last_processed_bitcoin_block: u32,
```
mpavlovic-txfusion, line 17: For now this is fine, but I believe when implementing the message processor we will need some kind of the serial number like `next_expected_message_id` to determine the next message to process.

PR #12, opened 2024-09-05
feat: `btc_watch` core

steph-mipt:


core/node/via_btc_watch/src/lib.rs
```diff
@@ -1,20 +1,26 @@
...
 
-use crate::message_processors::MessageProcessorError;
-
-mod message_processors;
+use self::message_processors::{L1ToL2MessageProcessor, MessageProcessor, MessageProcessorError};
 
-// re-exporting here isn't necessary, but it's a good practice to keep all the public types in one place
-pub use via_btc_client::types::BitcoinNetwork;
+// Number of blocks that we should wait before processing the new blocks.
+pub(crate) const BTC_BLOCKS_LAG: u32 = 1000;
```
irnb, line 17: can we read it from env variables?

and 1000 is for testing? right?

core/node/via_btc_watch/src/lib.rs
```diff
@@ -1,20 +1,26 @@
...
 
 #[derive(Debug)]
 struct BtcWatchState {
     last_processed_bitcoin_block: u32,
+    next_expected_priority_id: PriorityOpId,
```
irnb, line 22: i saw you created the next_expected_priority_id based on transaction serial id and vector length


could you please explain a little bit about the calculation and storing of the  `PriorityOpId`?

it's dependent on memory or you reading some data from db?
what is serial_id and how it's related to `PriorityOpId`

irnb (approved):

core/node/via_btc_watch/src/lib.rs
steph-mipt: basically it's the same thing. unique id for each transaction (check the transaction dal for more info)

PR #22, opened 2024-09-30
feat: Integrate bitcoin functionality

steph-mipt:
## Major Bitcoin Integration

This pull request introduces significant enhancements to our Bitcoin-related functionality and implements several new features:

* :rocket: Implemented `btc_watch` layer
* :moneybag: Implemented `btc_sender` layer
* :wrench: Applied critical fixes and updates to `inscriber` & `indexer` components
* :computer: Added basic implementation of `via` CLI
* :test_tube: Introduced new unit tests for Bitcoin-related modules
* :card_file_box: Created new database migrations
* :gear: Improved configuration creation and handling
* :package: Established `via-server` crate with essential layers (including `btc_watch` and `btc_sender` related components)


core/lib/via_btc_client/src/indexer/mod.rs
```diff
@@ -204,28 +229,28 @@ impl BitcoinInscriptionIndexer {
     fn is_valid_message(&self, message: &FullInscriptionMessage) -> bool {
         match message {
             FullInscriptionMessage::ProposeSequencer(m) => {
-                let is_valid = Self::get_sender_address(&m.common, self.network)
-                    .map_or(false, |addr| self.verifier_addresses.contains(&addr));
-                debug!("ProposeSequencer message validity: {}", is_valid);
-                is_valid
+                // let is_valid = parser::get_btc_address(&m.common, self.network)
```
irnb, line 232: @steph-mipt Let's create tasks for this part

and fix it as we discussed in slack ( getting the p2wpkh address from fee payer utxo sender instead of compress public key inside of inscription message.

core/lib/via_btc_client/src/indexer/parser.rs
```diff
@@ -452,6 +507,31 @@ fn find_via_inscription_protocol(instructions: &[Instruction]) -> Option<usize>
     position
 }
 
+pub fn get_btc_address(common_fields: &CommonFields, network: Network) -> Option<Address> {
+    let internal_pubkey =
+        UntweakedPublicKey::from_slice(&common_fields.encoded_public_key.as_bytes()).ok()?;
```
irnb, line 512: `&common_fields.encoded_public_key.as_bytes()`

.as_bytes() itself return reference so we can remove & from this input.

irnb (approved):
LGTM 🔥

etc/env/configs/via.toml
```diff
@@ -0,0 +1,63 @@
...
+btc_node_poll_interval=300
+rpc_url = "http://0.0.0.0:18443"
+rpc_user = "rpcuser"
+rpc_password = "rpcpassword"
+network = "regtest"
+bootstrap_txids = []
+actor_role = "Sequencer"
+
+[via_btc_sender]
+poll_interval = 1
```
idirall22, line 14: This is the time required to restart an execution cycle in btc_sender tasks. The value is in milliseconds, let's set a bigger value, I think for testing 30sec (30000) is good.

core/node/node_framework/src/implementations/layers/via_btc_watch.rs
```diff
@@ -0,0 +1,91 @@
...
+            .collect::<Result<Vec<_>, _>>()?;
+        let btc_watch = BtcWatch::new(
+            self.btc_watch_config.rpc_url(),
+            network,
+            node_auth,
+            self.btc_watch_config.confirmations_for_btc_msg,
+            bootstrap_txids,
+            main_pool,
+            self.btc_watch_config.poll_interval(),
+        )
```
idirall22, line 75: What do you think about passing the btc_watch_config to the BtcWatch instead of splitting it into multiple function parameters?

etc/env/configs/via.toml
```diff
@@ -0,0 +1,63 @@
+__imports__ = [ "base", "l2-inits/dev.init.env" ]
+
+[via_btc_watch]
+confirmations_for_btc_msg=3
+btc_node_poll_interval=300
```
idirall22, line 5: This value is in milliseconds, we should increase it to higher value, wdyt?

core/bin/via_server/src/config.rs
```diff
@@ -0,0 +1,94 @@
...
+    })
+}
+
+// TODO: temporary solution, should be removed after the config is refactored
+pub(crate) fn via_load_env_config(
+) -> anyhow::Result<(ViaBtcWatchConfig, ViaBtcSenderConfig, ViaCelestiaConfig)> {
+    let btc_watch_config =
+        ViaBtcWatchConfig::from_env().context("Failed to load BTC watch config")?;
+    let btc_sender_config =
+        ViaBtcSenderConfig::from_env().context("Failed to load celestia config")?;
```
mpavlovic-txfusion, line 89: Small copy/paste error, but can be misleading

idirall22 (changes requested):

etc/env/configs/via.toml
```diff
@@ -0,0 +1,63 @@
...
+network = "regtest"
+bootstrap_txids = []
+actor_role = "Sequencer"
+
+[via_btc_sender]
+poll_interval = 1
+private_key = "cVZduZu265sWeAqFYygoDEE1FZ7wV9rpW5qdqjRkUehjaUMWLT1R"
+max_aggregated_blocks_to_commit = 10
+max_aggregated_proofs_to_commit = 10
+max_txs_in_flight = 10
```
0xatomFusion, line 18: Let's set the btc_sender config for dev-net to:
```
max_aggregated_blocks_to_commit = 1
max_aggregated_proofs_to_commit = 1
max_txs_in_flight = 1
```
This means we will process onel1_batch at a time.

etc/env/configs/via.toml
0xatomFusion, line 14: This is the time required to restart an execution cycle in btc_sender tasks. The value is in milliseconds, let's set a bigger value, I think for testing 30sec (30000) is good.

etc/env/configs/via.toml
0xatomFusion, line 5: This is the time required to restart an execution cycle in btc_watch tasks. The value is in milliseconds, let's set a bigger value.

core/node/node_framework/src/implementations/layers/via_btc_watch.rs
```diff
@@ -0,0 +1,91 @@
...
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+        let btc_watch = BtcWatch::new(
+            self.btc_watch_config.rpc_url(),
+            network,
+            node_auth,
+            self.btc_watch_config.confirmations_for_btc_msg,
+            bootstrap_txids,
+            main_pool,
+            self.btc_watch_config.poll_interval(),
```
0xatomFusion, line 74: [Suggestion] What do you think about passing the btc_watch_config to the BtcWatch instead of splitting it into multiple function parameters?

0xatomFusion (changes requested):

etc/env/configs/via.toml
steph-mipt: Let's make it to 1 sec. As we have new blocks each 5 seconds or so, it should be fine
but good catch

core/node/node_framework/src/implementations/layers/via_btc_watch.rs
steph-mipt: let's keep the args as it is. probably we will change configs for watcher. when we do that, we will have explicit problem here, so we could fix it really fast

0xatomFusion (approved):

mpavlovic-txfusion (approved):

PR #27, opened 2024-10-04
feat: verifier network mock

mpavlovic-txfusion:
## Why ❔

- Provides mocked functionality of the Verifier Network needed for the Devnet 1 phase.

## Checklist

- [x] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [x] Code has been formatted via `zk fmt` and `zk lint`.

### DIFF
--- a/core/lib/via_btc_client/examples/deposit.rs
+++ b/core/lib/via_btc_client/examples/deposit.rs
@@ -1,6 +1,7 @@
+use std::env;
+
 use anyhow::{Context, Result};
 use bitcoin::{address::NetworkUnchecked, Amount, Txid};
-use std::env;
 use tracing::info;
 use via_btc_client::{
     inscriber::Inscriber,
--- a/core/lib/via_btc_client/examples/deposit.rs
+++ b/core/lib/via_btc_client/examples/deposit.rs
@@ -25,8 +25,6 @@ async fn main() -> Result<()> {
 
     let depositor_private_key = "cVZduZu265sWeAqFYygoDEE1FZ7wV9rpW5qdqjRkUehjaUMWLT1R".to_string();
 
-    let receiver_l2_address = EVMAddress::from_str("0x36615Cf349d7F6344891B1e7CA7C72883F5dc049")?;
-
     let bridge_p2wpkh_mpc_address = "bcrt1qdrzjq2mwlhrnhan94em5sl032zd95m73ud8ddw"
         .parse::<BitcoinAddress<NetworkUnchecked>>()?
         .require_network(NETWORK)?;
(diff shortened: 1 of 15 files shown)


core/lib/via_btc_client/src/inscriber/mod.rs
```diff
@@ -98,6 +98,7 @@ impl Inscriber {
         &mut self,
         input: &InscriptionMessage,
         config: InscriptionConfig,
+        recipient: Option<Recipient>,
```
0xatomFusion, line 101: What about if we move this inside the `InscriptionConfig` struct?

0xatomFusion (commented):
Good job!

core/lib/via_btc_client/src/indexer/mod.rs
```diff
@@ -318,16 +318,26 @@ impl BitcoinInscriptionIndexer {
...
             FullInscriptionMessage::ValidatorAttestation(va) => {
-                debug!("Processing ValidatorAttestation message");
-                if state.proposed_sequencer.is_some() {
-                    // warn!("sender_address : {:?}", sender_address);
-                    if let Some(proposed_txid) = state.proposed_sequencer_txid {
-                        if va.input.reference_txid == proposed_txid {
-                            state
-                                .sequencer_votes
-                                .insert(state.verifier_addresses[0].clone(), va.input.attestation);
+                if let Some(sender_address) = parser::get_btc_address(&va.common, network) {
```
steph-rs, line 321: so let's remove this check since `get_btc_address` is wrong

core/lib/via_btc_client/examples/deposit.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    info!(
+        "Depositor L1 balance: {}",
+        inscriber
+            .get_balance()
+            .await
+            .context("Failed to get balance")?
+    );
+
+    let receiver_l2_address = EVMAddress::random();
```
steph-rs, line 54: don't we want to deposit to some address? why is it random?

steph-rs (changes requested):

core/lib/via_btc_client/src/inscriber/mod.rs
mpavlovic-txfusion: We could, but since it's a BTC transfer that means it's not actually related to the inscription itself...

core/lib/via_btc_client/src/indexer/mod.rs
mpavlovic-txfusion: Actually I wanted to check here if `sender_address` exists (even if it is not calculated correctly by the `get_btc_address`), because we need to populate the `sequencer_votes` map (line 332) otherwise the bootstrap process won't  be completed.

core/lib/via_btc_client/examples/deposit.rs
mpavlovic-txfusion: Good one, I added one of the EVM rich wallet addresses with the known private key

steph-rs (approved):

irnb (approved):

PR #30, opened 2024-10-08
Feat/via state keeper

irnb:
state keeper layer


irnb:
the `via_node_sync` will be remove and instead of that i'm working to implement `StateKeeperIO`  for extrnalIO based on the bitcoin limitation or mock it, i'm not sure yet.

mpavlovic-txfusion (approved):

steph-rs (approved):

PR #34, opened 2024-10-10
Fix/ Code Panic During the genesis phase

irnb:
**Fixing the Issue with Code Panic During the First `via server` Execution**

**Approach:**
I’m adopting a similar approach to what zkSync uses. When running the server for the first time, there are two steps 
required:
1. `via server --genesis`
2. `via server`

However, if the **genesis block already exists in the database**, you only need to run:
• `via server`

### DIFF
--- a/core/bin/via_server/src/node_builder.rs
+++ b/core/bin/via_server/src/node_builder.rs
@@ -358,6 +358,16 @@ impl ViaNodeBuilder {
         Ok(self)
     }
 
+    /// Builds the node with the genesis initialization task only.
+    pub fn only_genesis(mut self) -> anyhow::Result<ZkStackService> {
+        self = self
+            .add_pools_layer()?
+            .add_query_eth_client_layer()?
+            .add_storage_initialization_layer(LayerKind::Task)?;
+
+        Ok(self.node.build())
+    }
+
     pub fn build(self) -> anyhow::Result<ZkStackService> {
         Ok(self
             .add_pools_layer()?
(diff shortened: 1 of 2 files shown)


core/bin/via_server/src/node_builder.rs
```diff
@@ -367,7 +377,7 @@ impl ViaNodeBuilder {
             .add_circuit_breaker_checker_layer()?
             .add_postgres_metrics_layer()?
             .add_query_eth_client_layer()?
-            .add_storage_initialization_layer(LayerKind::Task)?
+            .add_storage_initialization_layer(LayerKind::Precondition)?
```
steph-rs-1, line 380: why do we need this layers twice here?

steph-rs, line 380: why do we need these layers twice here?

steph-rs (approved):

core/bin/via_server/src/node_builder.rs
irnb: 1. The precondition version (called earlier in the build process) ensures that the storage is in a valid state before other components of the node are initialized. This might involve checking for the presence of genesis data or other critical information.
2- The task version (called later in the build process) might perform more extensive initialization or ongoing maintenance of the storage. This could include tasks like applying updates, performing migrations, or initializing components that depend on a fully set up storage system.

and also this is the same behavior that the zksync node builder apply to initializing their node.

PR #35, opened 2024-10-10
Feat/commit and proof flow

irnb:
## What ❔

adding layers are required for whole commitment and proof flow

we should first merge #34 before this branch


core/bin/via_server/src/node_builder.rs
```diff
@@ -358,6 +359,42 @@ impl ViaNodeBuilder {
...
+        self.node.add_layer(CommitmentGeneratorLayer::new(
+            self.genesis_config.l1_batch_commit_data_generator_mode,
+        ));
+
+        Ok(self)
+    }
+
+    fn add_da_dispatcher_layer(mut self) -> anyhow::Result<Self> {
+        let eth_sender_config = try_load_config!(self.configs.eth);
+        if let Some(sender_config) = eth_sender_config.sender {
```
mpavlovic-txfusion, line 372: Ok for now, but lets not forget to remove `EthConfig` usage and move the required config to some Via related one.

irnb: actually we don't need this check, cause we only have validium mode in btc_sender, so we can remove this check.

core/node/via_btc_sender/src/btc_inscription_aggregator.rs
```diff
@@ -128,13 +128,13 @@ impl ViaBtcInscriptionAggregator {
 
     // Todo: call indexer to fetch  the data
     async fn get_bootloader_code_hash(&self) -> anyhow::Result<H256> {
-        let hex_str = "0000000000000000000000000000000000000000000000000000000000000000";
+        let hex_str = "010008e742608b21bf7eb23c1a9d0602047e3618b464c9b59c0fba3b3d7ab66e";
```
steph-rs, line 131: so it will remain constant, right?

migrations/20241016074609_get_ready_for_dummy_proof_dispatch_l1_batches.sql
```diff
@@ -0,0 +1 @@
+-- Add migration script here
```
steph-rs, line 1: there's no migration, is it right?

steph-rs (approved):

core/node/via_btc_sender/src/btc_inscription_aggregator.rs
irnb: yes, for this phase i just keep it constant, but for next phases i will change this to use correct db table.

migrations/20241016074609_get_ready_for_dummy_proof_dispatch_l1_batches.sql
irnb: Oh, thanks for mentioning it
I removed it

mpavlovic-txfusion (approved):

infrastructure/via/src/celestia.ts
```diff
@@ -0,0 +1,159 @@
...
+        await runCommand(restart_celestia_container_command);
+    } catch (error) {
+        console.error('Error executing command:', error);
+    }
+
+    try {
+        await get_celestia_faucet_token(node_address);
+        await get_celestia_faucet_token(node_address);
+        console.log('Request Sent to Faucet');
+        console.log(`Check your balance at https://arabica.celenium.io/address/${node_address}?tab=transactions`);
```
steph-rs, line 153: should we remove this? because now we are using local setup

PR #50, opened 2024-10-22
feat: metrics

steph-rs:
This PR provides new `via_server_btc_watch` metrics. The list of all edited metrics (prefixes) is:
| zksync    | via |
| -------- | ------- |
| server_da_dispatcher | via_server_da_dispatcher |
| server_state_keeper |  via_server_state_keeper |
| server_tx_aggregation | via_server_tx_aggregation |
| server_state_keeper_l1_batch | via_server_state_keeper_l1_batch |
| server_state_keeper_miniblock | via_server_state_keeper_miniblock |
| batch_tip | via_batch_tip |
| server_state_keeper_updates_manager | via_server_state_keeper_updates_manager |
| server_eth_watch | via_server_btc_watch |

### DIFF
--- a/core/node/via_btc_watch/src/lib.rs
+++ b/core/node/via_btc_watch/src/lib.rs
@@ -18,6 +18,7 @@ use self::{
     message_processors::{L1ToL2MessageProcessor, MessageProcessor, MessageProcessorError},
     metrics::METRICS,
 };
+use crate::metrics::ErrorType;
 
 // Number of blocks that we should wait before processing the new blocks.
 pub(crate) const BTC_BLOCKS_LAG: u32 = 1000;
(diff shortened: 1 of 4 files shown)


0xatomFusion (commented):
It will be nice to add metrics related to errors, like network failures, number of errors happens in the layer. Also, it could be interesting to add counter for inscriptions processed.
@steph-rs wdyt?

0xatomFusion (commented):
Another thing I forgot to add in last comment is that we can improve the metrics by adding lables.
For example, we can have something like this:

```
METRICS.inscriptions_processed(&["deposit"]).inc();
METRICS.inscriptions_processed(&["verify"]).inc();
...
```
The idea is to use the same metric but with different lables to have more details about what happens during the execution. Like this we can build a good alerts system.

We can do same for errors by adding lables for ["network error", "execution error", ...]

0xatomFusion (approved):

PR #67, opened 2024-12-03
Feat/transaction builder

irnb:
**via_musig2 Wrapper**
**Transaction Builder Component** for withdrawal
**Bridge Address Calculation:** Adds new command to via 

**Coordinator Example:** Demonstrates pooling for normal verifiers and an HTTP server for the coordinator. Actual implementation to be done in the withdrawal service.


core/lib/via_btc_client/src/withdrawal/mod.rs
mpavlovic-txfusion, line 1: Maybe better to rename this folder into withdrawal_builder, just to keep the same logic as for other components of the via_btc_client.

core/lib/via_musig2/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+
+impl Signer {
+    /// Create a new signer with the given secret key and index
+    pub fn new(
+        secret_key: SecretKey,
+        signer_index: usize,
+        all_pubkeys: Vec<PublicKey>,
+    ) -> Result<Self, MusigError> {
+        let secp = Secp256k1::new();
+        let public_key = PublicKey::from_secret_key(&secp, &secret_key);
```
mpavlovic-txfusion, line 55: Maybe we can check here that the computed `public_key` is indeed passed in `all_pubkeys` at `signer_index`?

irnb: I think it's better to doing this in withdrawal service, during the initialization of service to prevent doing the same check for every signing processes

wdyt?

PR #68, opened 2024-12-05
feat: upgrade the protocol to bitcoin version

0xatomFusion:
## What ❔

- Upgrade the bootloader binary files and bump the protocol version to Bitcoin 26.

## Why ❔

Those modifications are required after the modifications added to this [PR](https://github.com/vianetwork/era-contracts/pull/3)

## Checklist


steph-rs:
do we need these changes to update the way how vm charges fees, right? or anything else?

0xatomFusion:
> do we need these changes to update the way how vm charges fees, right? or anything else?
The VM fees logic was not modified. But after changing the L2BaseToken contract in this [PR](https://github.com/vianetwork/era-contracts/pull/3) , the new binary hashes where generated.
And if we don't upgrade, the bootloader fails in this [condition](https://github.com/vianetwork/era-contracts/blob/main/system-contracts/bootloader/bootloader.yul#L746C41-L746C68)

steph-rs (approved):

irnb:
https://github.com/vianetwork/era-contracts/pull/3#discussion_r1872973892

etc/env/file_based/genesis.yaml
```diff
@@ -1,11 +1,11 @@
-genesis_root: 0x01afc682f1b12cd5bcdacd0e04505ffdd4aafe459e44150c422a86437ae13e4e
+genesis_root: 0xa8ba1e76b4d8560bee8c039c91d88b4860cae79b05d4b2d656f0d38d7856a2dd
 genesis_rollup_leaf_index: 54
-genesis_batch_commitment: 0xe3ace05ee94258b46086c437763f8c669c58b2a24f11f8b719cf2efaec9daaed
+genesis_batch_commitment: 0x53de9f8fb086c1cc78e337800264fd9beb7a35fc56dd4a7afce5b3e9f2475251
 genesis_protocol_semantic_version: '0.24.1'
 # deprecated
 genesis_protocol_version: 24
```
irnb, line 6: Should we update the `genesis_protocol_semantic_version` to 26 to enable Bitcoin from genesis, or it's not necessary?

0xatomFusion: Fixed

PR #70, opened 2024-12-05
Feat/Start celestia node from last block

0xatomFusion:
## What ❔

Start Celestia node from the last minted block (current block).

## Why ❔

Before, the Celestia node needed to sync from the first block with the remote node before be able to query data from it.

## Checklist

- [] Documentation comments have been added / updated.

### DIFF
--- a/infrastructure/via/src/config.ts
+++ b/infrastructure/via/src/config.ts
@@ -224,4 +224,3 @@ command
         diff = diff ? diff : '0';
         pushConfig(environment, diff);
     });
-// .action(fetchCelestiaTrustedHash);
--- a/infrastructure/via/src/config.ts
+++ b/infrastructure/via/src/config.ts
@@ -6,6 +6,7 @@ import * as env from './env';
 import path from 'path';
 import dotenv from 'dotenv';
 import { unpackStringSemVer } from 'utils';
+import { updateEnvVariable } from './helpers';
 
 function loadConfigFile(configPath: string, stack: string[] = []) {
     if (stack.includes(configPath)) {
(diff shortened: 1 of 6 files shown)


infrastructure/via/src/config.ts
```diff
@@ -207,3 +224,4 @@ command
         diff = diff ? diff : '0';
         pushConfig(environment, diff);
     });
+// .action(fetchCelestiaTrustedHash);
```
steph-rs, line 227: it's commented, is it right?

steph-rs (approved):

infrastructure/via/src/config.ts
0xatomFusion: I need to clean this

mpavlovic-txfusion (approved):

infrastructure/via/src/config.ts
```diff
@@ -180,8 +181,24 @@ export function pushConfig(environment?: string, diff?: string) {
     env.modify('DATABASE_MERKLE_TREE_BACKUP_PATH', `./db/${environment}/backups`, l2InitFile, false);
 
     env.reload();
+    await fetchCelestiaTrustedHash();
 }
 
+const fetchCelestiaTrustedHash = async () => {
```
mpavlovic-txfusion, line 187: Maybe better to make this part of `via celestia` command, since we already have a logic there for getting the celestia auth_token and updating configs?

infrastructure/via/src/helpers.ts
```diff
@@ -0,0 +1,16 @@
+import * as fs from 'fs/promises';
+import * as dotenv from 'dotenv';
+
+export async function updateEnvVariable(envFilePath: string, variableName: string, newValue: string) {
```
mpavlovic-txfusion, line 4: This is also already present in `celestia.ts`. It makes sense to me to be a helper function, just reuse it then in celestia.ts

infrastructure/via/src/config.ts
0xatomFusion: We need to fetch the celestia Block hash before starting the containers because the block hash is injected through ENV. This needs to be done before running `make init`.

infrastructure/via/src/helpers.ts
0xatomFusion: Fixed

irnb (approved):

PR #77, opened 2024-12-20
Feat/btc sender voting inscription

0xatomFusion:
## What ❔
- Implement the verifier vote inscription node and layer.
- Implement e2e tests for vote inscription.

## Why ❔
- The verifier should be able to vote when a batch is verified.

## Testing
1. Run `make env`
2. Run `make config`
3. Run `make init`
4. Run the tests `cargo test --package via_btc_sender --  --nocapture`

Screenshot from 2024-12-20 18-04-29

## Checklist


core/lib/dal/src/via_votes_dal.rs
```diff
@@ -165,4 +165,82 @@ impl ViaVotesDal<'_, '_> {
 
         Ok(row.max_batch_number.map(|n| n as u32))
     }
+
+    pub async fn verify_votable_transaction(
```
steph-rs, line 169: what's the difference between these two columns? finilized & verified? could you explain, please?

steph-rs: nvm, got it

steph-rs (commented):
let's add usage of `actor_role` from `ViaBtcSenderConfig`. and maybe try to add new layer to new bin?

0xatomFusion:
> let's add usage of `actor_role` from `ViaBtcSenderConfig`. and maybe try to add new layer to new bin?

We don't need to add a role to the config because, I separated the verifier and the sequencer inscriptions logic into different layers. So based on the orchestration layer we can select the layer we want to use.
Yes, I forgot to add the layer to the verifier bin

core/lib/dal/src/via_verifier_blocks_dal.rs
```diff
@@ -0,0 +1,79 @@
...
+                );
+                let result = instrumentation
+                    .clone()
+                    .with(query)
+                    .execute(self.storage)
+                    .await?;
+
+                if result.rows_affected() == 0 {
+                    let err = instrumentation.constraint_error(anyhow::anyhow!(
+                        "Update commit_l1_batch_inscription_id that is is not null is not allowed"
```
mpavlovic-txfusion, line 50: Wrong column name

core/lib/dal/src/via_verifier_blocks_dal.rs
```diff
@@ -0,0 +1,79 @@
...
+            r#"
+            SELECT EXISTS(
+                SELECT 1
+                FROM via_l1_batch_vote_inscription_request
+                WHERE l1_batch_number = $1
+            )
+            "#,
+            batch_number
+        )
+        .instrument("check_vote_l1_batch_inscription_request_id_exists")
```
mpavlovic-txfusion, line 73: Assigned name doesn't match the function name

core/lib/dal/src/via_votes_dal.rs
```diff
@@ -165,4 +165,82 @@ impl ViaVotesDal<'_, '_> {
...
+            tx_id.as_bytes(),
+            l1_batch_status
+        )
+        .instrument("verify_transaction")
+        .execute(self.storage)
+        .await?;
+        Ok(())
+    }
+
+    pub async fn get_first_not_finilized_block(&mut self) -> DalResult<Option<i64>> {
```
mpavlovic-txfusion, line 196: typo

core/lib/dal/src/via_votes_dal.rs
```diff
@@ -165,4 +165,82 @@ impl ViaVotesDal<'_, '_> {
...
+        let l1_block_number = sqlx::query_scalar!(
+            r#"
+            SELECT
```
mpavlovic-txfusion, line 199: Maybe a more optimised and intuitive query would be:

`SELECT
    MIN(l1_batch_number) as "l1_batch_number"
  FROM
    via_votable_transactions
  WHERE
    is_finalized = FALSE
`

core/lib/dal/migrations/20241209150000_create_via_votes.up.sql
```diff
@@ -2,6 +2,8 @@ CREATE TABLE IF NOT EXISTS via_votable_transactions (
     l1_batch_number BIGINT NOT NULL REFERENCES l1_batches (number) ON DELETE CASCADE,
     tx_id BYTEA,
     is_finalized BOOLEAN NOT NULL DEFAULT FALSE,
+    is_verified BOOLEAN NOT NULL DEFAULT FALSE,
```
mpavlovic-txfusion, line 5: Can you explain the purpose of `is_verified` and `l1_batch_status`?

0xatomFusion: The `is_verified` is updated by the verifier after the ZKP is done and the `l1_batch_status` is the output of the verification valid or not valid.

core/lib/dal/src/via_verifier_blocks_dal.rs
0xatomFusion: fixed

core/lib/dal/src/via_votes_dal.rs
0xatomFusion: fixed

core/lib/dal/src/via_votes_dal.rs
0xatomFusion: fixed

core/lib/dal/migrations/20241209150000_create_via_votes.up.sql
mpavlovic-txfusion: I don't think i fully get the idea here. `via_votable_transactions` table was previously used for the voting mechanism (meaning `btc_watch` Sequencer component) and here it looks that we are using the same table for the Verifier db to store the verification status for this one Verifier.
Can you elaborate on this?

0xatomFusion: The `via_votable_transactions`  stores the l1_block data that is ready to be verified. 
### Flow

1. The `btc_watch` detects that there is a new `proof inscription`, it inserts a new row in the `via_votable_transactions`.
2. The verifier task detects that there is a new votable transaction that was not verrified (`is_verified=FALSE`).
3. The verifier starts the ZKP verification. After finishing it update the `via_votable_transactions` columns `l1_batch_status` (if valid proof set to `TRUE` else `FALSE`) and `is_verified=TRUE`.
4. The btc_sender detects that the verification is finished and it needs to create a voting inscription.
5. The voting inscription is created and sent.
6. The l1_batch is finailized

The `via_votable_transactions` should be migrated to the verifer DAL in another PR as it's related to verification.

steph-rs (approved):

mpavlovic-txfusion (approved):

PR #87, opened 2025-01-12
chore: new btc_watch for verifier, verifier related crates moved

steph-rs:
## Checklist

- [ ] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [ ] Code has been formatted via `zk fmt` and `zk lint`.


via_verifier/node/via_btc_watch/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+
+        let message_processors: Vec<Box<dyn MessageProcessor>> = vec![Box::new(
+            VerifierMessageProcessor::new(DEFAULT_VOTING_THRESHOLD),
+        )];
+
+        let confirmations_for_btc_msg = confirmations_for_btc_msg.unwrap_or(0);
+
+        // We should not set confirmations_for_btc_msg to 0 for mainnet,
+        // because we need to wait for some confirmations to be sure that the transaction is included in a block.
+        if network == BitcoinNetwork::Bitcoin && confirmations_for_btc_msg == 0 {
```
0xatomFusion, line 71: Is it possible to move this ENV validation to the config crate?

via_verifier/node/via_btc_watch/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+            match storage.via_votes_dal().get_last_inserted_block().await? {
+                Some(block) => block.saturating_sub(1),
+                None => indexer
+                    .fetch_block_height()
+                    .await
+                    .context("cannot get current Bitcoin block")?
+                    .saturating_sub(btc_blocks_lag as u128) as u32, // TODO: remove cast
+            };
+
+        // TODO: get the bridge address from the database?
```
0xatomFusion, line 103: The bridge address needs to be computed using all verifiers pubkeys, so I  think the best way is to get it from ENVs.

0xatomFusion (approved):
Great PR!!
I left some minor comments :)

irnb (approved):

via_verifier/node/via_btc_watch/src/lib.rs
steph-rs: I think it will be good to have this logic here, when we want to run the via on btc mainnet to find this here to change it

mpavlovic-txfusion (approved):

PR #104, opened 2025-01-22
Feat/via add verify unsigned tx to verifier

0xatomFusion:
## What ❔

Add logic to verify the unsigned transaction by the verifier before sending starting the signing session.
The verifier checks:
1. If the withdrawals are included in the `tx_out` list.
2. If the OP_RETURN code includes the valid `proof_tx_id`.
3. If the `message_to_sign` is valid.

This PR doesn't validate if the if the fee paid to the minter is not over estimated.

## Testing
1. Run `make all`
2. Run `make via-coordinator`
3. Run  `make via-verifier`
4. Deposit 10 BTC
5. Withdraw 2 times fast to get the both transactions included in the same l1 batch (you can increase the sealing time)

Screenshot from 2025-01-27 09-05-39

Above you can see that I requested a total of 5 BTC
The transaction should be verified and processed.

Screenshot from 2025-01-27 09-05-03

Done :)
 
## Checklist


via_verifier/node/withdrawal_service/src/verifier/mod.rs
```diff
@@ -135,6 +149,101 @@ impl ViaWithdrawalVerifier {
...
+
+        let len = withdrawals.len();
+        if len == 0 {
+            tracing::error!(
+                "Invalid session, there are no withdrawals to process, l1 batch: {}",
+                session.l1_block_number
+            );
+            return Ok(false);
+        }
+        if len + 2 != unsigned_tx.tx.output.len() {
```
mpavlovic-txfusion, line 190: I think we should adjust these checks, since we group the withdrawals by address in order to have smaller withdrawal tx with fewer outputs (check this logic https://github.com/vianetwork/via-core/blob/d0421359b2482440a96006388729c97f5179a6d7/core/lib/via_btc_client/src/withdrawal_builder/mod.rs#L84).

0xatomFusion: Fixed in this [commit](https://github.com/vianetwork/via-core/pull/104/commits/c2636558b167f5182981551019bd7616a1ede769)

mpavlovic-txfusion: Not sure this fixes it.
My point was that, for example, if we have 3 withdrawals to the same btc address that will produce only 1 output in the `unsigned_tx.tx.output` with the sum on the amounts. If am I right here, then the validation should not return false if  `len + 2 != unsigned_tx.tx.output.len()`.
Also the amounts can not be checked in the way we check it at line 204.

0xatomFusion: Fixed [commit](https://github.com/vianetwork/via-core/pull/104/commits/ea0892c8d443d27c6e616422578531f5377fd6ab)

steph-rs (approved):

mpavlovic-txfusion (approved):

PR #105, opened 2025-01-23
feat(logic): add first step filter for transaction in btc block

irnb:
WIP


0xatomFusion (approved):
Great job!! Bridge with OP return works fine!

PR #106, opened 2025-01-23
Fix/via verifier session managment

0xatomFusion:
## What ❔
- Start a new session after the previous signing session was finished and the transaction minted in a block.
- Set the minimum fee to 1 (sometimes the Regtest returns 0).

## Why ❔
The issue is when the session is finished all the coordinator broadcast the signed transaction, it get stuck to start a new session.

## Testing
1. Run the sequencer
2. Deposit 100 BTC
3. withdraw 5 BTC
4. withdraw 3 BTC
5. Deposit 10 BTC
6. Start the coordinator
7. Start the verifier
8. Check on BTC explorer if all withdrawals were processed.
9. Check the database data.

Screenshot from 2025-01-23 15-20-12

## Checklist

### DIFF
--- a/via_verifier/node/withdrawal_service/src/verifier/mod.rs
+++ b/via_verifier/node/withdrawal_service/src/verifier/mod.rs
@@ -1,7 +1,7 @@
 use std::{collections::HashMap, sync::Arc};
 
 use anyhow::{Context, Result};
-use bitcoin::{hashes::Hash, TapSighashType, Witness};
+use bitcoin::{hashes::Hash, TapSighashType, Txid, Witness};
 use musig2::{CompactSignature, PartialSignature};
 use reqwest::{header, Client, StatusCode};
 use tokio::sync::watch;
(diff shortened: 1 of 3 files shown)


via_verifier/node/withdrawal_service/src/verifier/mod.rs
```diff
@@ -73,36 +73,65 @@ impl ViaWithdrawalVerifier {
...
     async fn loop_iteration(&mut self) -> Result<(), anyhow::Error> {
-        if self.config.verifier_mode == VerifierMode::COORDINATOR {
-            self.create_new_session().await?;
+        let mut session_info = self.get_session().await?;
 
+        if self.config.verifier_mode == VerifierMode::COORDINATOR {
             tracing::info!("create a new session");
-            self.build_and_broadcast_final_transaction().await?;
+
+            if !session_info.l1_block_number == 0 {
```
steph-rs, line 81: I think you meant `session_info.l1_block_number != 0`

steph-rs: it doesn't matter but let's keep code readable

steph-rs (approved):

mpavlovic-txfusion (approved):

via_verifier/node/withdrawal_service/src/verifier/mod.rs
0xatomFusion: Fixed

PR #114, opened 2025-02-06
remove: retry logic from btc_sender

0xatomFusion:
## What ❔
This PR removes the redundant retry logic from the BTC sender component since transaction retry handling is now managed through the Inscriber context using fee increase mechanisms for stuck mempool transactions. This change aligns with our chaining transactions approach.

### Changes
- Remove retry logic from BTC sender.
- Clean up related retry configuration and tests.

## Why ❔
The current retry logic in the BTC sender has become redundant as we've implemented a more robust solution through the Inscriber context, which handles stuck transactions by increasing fees when necessary. This change:

### Testing
Screenshot from 2025-02-06 07-13-55

## Checklist


irnb (approved):

irnb:
Looks good to me. 🚀

However, let’s remember that the current architecture relies on a single Bitcoin node, which can become a single point of failure as we scale. In the future, it would be wise to add the capability to broadcast transactions to multiple Bitcoin nodes, and also fetch data from them if one node fails.

Since this PR also touches on handling failed inscriptions, I just wanted to note this for a future improvement. 

For now, let’s proceed and merge this PR. :grin:

PR #119, opened 2025-02-10
Feat/via verifier validate op priority

0xatomFusion:
## What ❔
The verifier indexes L1-to-L2 deposits and stores them in a local table called via_transactions. Before performing proof verification, the verifier first validates the op_priority_id by ensuring that deposits were processed sequentially. It then compares the bootloader's canonical transaction hash with its locally stored hash. If the values match, it confirms that the correct amount was minted and transferred to the intended recipient.

- Index the L1 to L2 deposits.
- Create a new via_da_client for the verifier and move the da logic from the withdrawal lib.
- Add transactions dal to store the deposits.
- Add logic to the zk verifier to validate the op_priority_id.

## Why ❔
The verifer network must check if the sequencer is processing the deposits sequentially and mint the correct amount of tokens on L2.

## Checklist


via_verifier/node/via_btc_watch/src/message_processors/l1_to_l2.rs
```diff
@@ -0,0 +1,205 @@
...
+                    MessageProcessorError::DatabaseError(e.to_string())
+                })?;
+        }
+
+        Ok(())
+    }
+}
+
+impl L1ToL2MessageProcessor {
+    fn create_l1_tx_from_message(
```
irnb, line 115: The `create_l1_tx_from_message` method is duplicated code from `core/node/via_btc_watch/src/message_processors/l1_to_l2.rs`. 

This duplication could lead to problems if changes are made to one copy but not the other. One suggestion is to extract the method into a shared library and have both services import it from there. Otherwise, we could add a `//NOTICE` comment at the top of both files, warning that any changes in one must be mirrored in the other.

via_verifier/node/via_zk_verifier/src/lib.rs
```diff
@@ -138,6 +150,62 @@ impl ViaVerifier {
...
+            .list_transactions_not_processed(deposit_logs.len() as i64)
+            .await?;
+
+        if txs.len() != deposit_logs.len() {
+            tracing::error!(
+                "Verifier did not index all the deposits, expected {} found {}",
+                deposit_logs.len(),
+                txs.len()
+            );
+            return Ok(false);
```
irnb, line 179: I don’t think this logic is necessary. We only need to prevent the sequencer from minting an asset based on a fake claim. It’s possible that we could have 100 pending deposits, but in a single L1 batch, the sequencer might only process 50 of them, and that’s fine.

We could consider changing the logic to check if there is a pending transaction waiting for multiple L1 batches and the sequencer hasn’t processed any deposit in a given L1 batch, then reject that batch. But again, this isn’t strictly necessary.

wdyt?

irnb (changes requested):

via_verifier/node/via_zk_verifier/src/lib.rs
```diff
@@ -138,6 +150,62 @@ impl ViaVerifier {
...
+        }
+
+        if txs.is_empty() {
+            tracing::error!("There is no transactions to validate the op priority id",);
+            return Ok(true);
+        }
+
+        for (i, raw_tx_id) in txs.iter().enumerate() {
+            let db_raw_tx_id = H256::from_slice(&raw_tx_id);
+            let log_raw_tx_id = deposit_logs[i].key;
```
irnb, line 189: This logic is also problematic. If we remove the `if txs.len() != deposit_logs.len()` check, the sequencer could include more deposits than we’re verifying, allowing it to insert a fake deposit at the end of the logs and potentially compromise the system.

To address this, we have two potential solutions:
1. Use a hash approach.
Iterate over all transactions based on the announced priority_id in the L1 batch. Then compute the `priority_queue_hash` and compare it to the hash provided in the L1 batch.
2. Iterate over `deposit_logs` instead of `txs`.
This way, we verify every deposit log directly, rather than relying on the transaction list alone.

via_verifier/node/via_zk_verifier/src/lib.rs
irnb: With this check in place, there’s also a risk of rejecting an honest sequencer’s L1 batch. 

For example, suppose the sequencer seals an L1 batch and waits for proof generation. Meanwhile, someone makes a deposit to the bridge, and the verifier detects this deposit transaction only after the batch is already sealed. This could lead to rejecting an otherwise valid L1 batch.

via_verifier/node/via_zk_verifier/src/lib.rs
```diff
@@ -138,6 +150,62 @@ impl ViaVerifier {
...
+                deposit_logs.push(log);
+            }
+        }
+
+        let txs = storage
+            .via_transactions_dal()
+            .list_transactions_not_processed(deposit_logs.len() as i64)
+            .await?;
+
+        if txs.len() != deposit_logs.len() {
```
irnb, line 173: With this check in place, there’s also a risk of rejecting an honest sequencer’s L1 batch.

For example, suppose the sequencer seals an L1 batch and waits for proof generation. Meanwhile, someone makes a deposit to the bridge, and the verifier detects this deposit transaction only after the batch is already sealed. This could lead to rejecting an otherwise valid L1 batch.

via_verifier/node/via_btc_watch/src/message_processors/l1_to_l2.rs
```diff
@@ -0,0 +1,205 @@
...
+        &self,
+        tx_id: H256,
+        serial_id: PriorityOpId,
+        msg: &L1ToL2Message,
+    ) -> Result<L1ToL2Transaction, MessageProcessorError> {
+        let amount = msg.amount.to_sat() as i64;
+        let eth_address_l2 = msg.input.receiver_l2_address;
+        let calldata = msg.input.call_data.clone();
+
+        let mantissa = U256::from(10_000_000_000u64); // Eth 18 decimals - BTC 8 decimals
```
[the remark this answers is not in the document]
0xatomFusion: This is the same code we use in the Sequencer side, I copy paste it but we have to refactor in another PR.
On L2 the BTC token is 18 decimals so we have to scale  the amount in satoshi to 10^10.

via_verifier/node/via_btc_watch/src/message_processors/l1_to_l2.rs
0xatomFusion: Yes, I did not want to refactor the Sequencer code in this PR to not mix and have a huge PR. I'll do it in a next PR, extract the common logic in a lib

via_verifier/lib/verifier_dal/migrations/20250207164238_via_add_transactions_dal.up.sql
```diff
@@ -0,0 +1,14 @@
+CREATE TABLE IF NOT EXISTS via_transactions (
+    "priority_id" BIGINT NOT NULL,
+    "tx_id" BYTEA NOT NULL,
+    "receiver" VARCHAR NOT NULL,
+    "value" BIGINT NOT NULL,
+    "calldata" BYTEA,
+    "canonical_tx_hash" BYTEA NOT NULL,
+    "status"  BOOLEAN,
```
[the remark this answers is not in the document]
0xatomFusion: The status represent the transaction execution status success or fails. On ZkSync this value is used on L1 to process refund transactions. For example, a user did not pay enough gas fee or other scenarios.
We this status we can implement the refund mechanism.

via_verifier/node/via_zk_verifier/src/lib.rs
0xatomFusion: I think we don't need to check the pending transactions as we can do it only when they are included in the batch.

via_verifier/node/via_zk_verifier/src/lib.rs
0xatomFusion: The verifier and Sequencer uses the same indexing L1ToL1Message logic, so we expect that they both will index the same transactions in the same order with same calldata.
Let's say:
1. we have 3 deposits, the sequencer and verifiers indexed them all.
2. The sequencer and verifier indexed them both correctly.
3. The Sequencer process only 2 in the batch 1.
4. The verifier decode the pubdata and check how much the sequencer processed, in this case he see 2.
5. The verifier select the 2 first transactions in order (ASC) using the priority id.
6. The verifier compare the `canonical_tx_hash` if the emitted one matches the indexed one. If correct it update the local table with the transaction status and return true (priority_id valid).
7. The sequencer include the 3rd deposit in batch 2.
8. The verifier does the same verification but this time for only the 3rd deposit...

If we want to implement the validation using the `priority_queue_hash`, we need to update the inscription to include it in the inscription. I investigated the solidity code and it's passed as argument when commit a new [batch](https://github.com/vianetwork/era-contracts/blob/00a881f98f69fad74a021040a5940cd6de9eefdc/l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol#L227-L228)

We are not relaying on the tx list alone, we compare the log with the tx indexed locally by the verifier.

via_verifier/node/via_zk_verifier/src/lib.rs
0xatomFusion: Can you check this [logic](https://github.com/vianetwork/via-core/blob/6fedb9d04265c84e55aaddd5af790a3cdb07dc1f/via_verifier/node/via_zk_verifier/src/lib.rs#L168-L172) related to txs selections?
We are selecting the txs using a LIMIT param that equal the logs. This check should not fail if the verifier is in sync with the network.

irnb: oh, my bad 

`deposit_logs.len()` we use the len of `deposit_logs`  as input for query and it's gonna fixed all issue in my mind

irnb (approved):

via_verifier/lib/verifier_dal/src/via_transactions_dal.rs
```diff
@@ -0,0 +1,126 @@
...
+            limit
+        )
+        .instrument("list_transactions")
+        .fetch_all(self.storage)
+        .await?;
+
+        let mut canonical_tx_hashs: Vec<Vec<u8>> = Vec::with_capacity(rows.len());
+        for row in rows {
+            canonical_tx_hashs.push(row.canonical_tx_hash);
+        }
```
mpavlovic-txfusion, line 79:
```suggestion
        let canonical_tx_hashs: Vec<Vec<u8>> = rows
            .into_iter()
            .map(|row| row.canonical_tx_hash)
            .collect();
```

via_verifier/node/via_zk_verifier/src/lib.rs
```diff
@@ -138,6 +150,62 @@ impl ViaVerifier {
...
+                let status = deposit_logs[i].value.is_zero();
+                storage
+                    .via_transactions_dal()
+                    .update_transaction(&log_raw_tx_id, !status)
+                    .await?;
+                continue;
+            }
+            tracing::error!("Sequencer did not process the deposit transactions in series for l1 batch {}, invalid priority id for transaction hash {}", l1_batch_number, db_raw_tx_id);
+            return Ok(false);
+        }
```
mpavlovic-txfusion, line 200: I would invest some time to try to make this more readable and prevent bugs. Maybe something like this can be used:

```
for (raw_tx_id, deposit_log) in txs.iter().zip(deposit_logs.iter()) {
    let db_raw_tx_id = H256::from_slice(raw_tx_id);
    if db_raw_tx_id != deposit_log.key {
        tracing::error!(
            "Sequencer did not process the deposit transactions in series for l1 batch {}, \
            invalid priority id for transaction hash {}", 
            l1_batch_number, 
            db_raw_tx_id
        );
        return Ok(false);
    }
    
    let status = !deposit_log.value.is_zero();
    storage
        .via_transactions_dal()
        .update_transaction(&deposit_log.key, status)
        .await?;
}
```

0xatomFusion: Fixed [Commit](https://github.com/vianetwork/via-core/pull/119/commits/5b523c54326cab964c74c392fb21cae6910cabe4)

mpavlovic-txfusion (approved):

PR #121, opened 2025-02-13
feat: Add via command for running unit tests

mpavlovic-txfusion:
## What ❔

- `via test rust` command added to via cli for running unit tests with `cargo-nextest`

## Why ❔

- Enables all Via tests (from all crates prefixed with `via_`) to be run with `via test rust -p 'via_*'`, even the ones from Via node components that need access to test db


0xatomFusion (approved):
Great job!

Screenshot from 2025-02-24 09-57-34

PR #123, opened 2025-02-14
Feat/enable stress test for via protocol

irnb:
Making LoadNext compatible with the Via Protocol specification.


irnb:
`via run loadtest`

PR #128, opened 2025-02-23
Feat/via block reverter

0xatomFusion:
## What ❔

This PR introduce the block reverter to the via network.
- Fix the decode attestation logic when the vote is false.
- Add the block reverter cli.
- Improve the database schema (core and verifier) to integrate the block revert logic with tests.
- Apply modification to the nodes impl.
- Add logic to the verifier and coordinator to simulate an invalid l1 batch and be able to trigger the revert process when **dev**.
- Delete the logic related to create a votable transaction from the sequencer because it's not used.
- Add logic to the DA dispatcher to stop processing data when it detects that there is an invalid l1 batch
Screenshot from 2025-02-22 13-26-40

### Sequencer

1. Admin can to roll back an L1 batch with not proof DA or invalid l1 batch from verifier network.
2. It's not possible to roll back a finalized l1 batch by the verifier network.
3. The Via DA dispatcher stop dispatch data if it detects that there is invalid l1 batch.

### Verifier network

1. The verifier will do a zk verification and vote if the l1 batch is valid or not.
2. Stops processing withdrawals from the invalid l1_batch.
3. The verifier waits for new l1 batch from the sequencer with a valid previous parent (create a new inscription chain
   fork).
4. The verifier deletes the invalid l1 batch and it's children when a new valid l1 batch is processed.
5. The verifier network continue processing l1 batches.

#### Case 1: the verifier network detects an invalid l1 batch

1. Start the Sequencer, verifier and the coordinator.
2. Update the [config](etc/env/configs/via_coordinator.toml) and set the l1 batch that we expect it's going to be
   invalid. For example, we expect that the l1 batch number `3` is invalid.
3. Deposit 10 BTC and wait for l1 batch to be created (`l1 batch 1`).
4. Withdraw 1 BTC and wait for l1 batch to be created (`l1 batch 2`).
5. Withdraw 1 BTC and wait for l1 batch to be created (`l1 batch 3`).
6. Withdraw 1 BTC and wait for l1 batch to be created (`l1 batch 4`).
7. The verifier network will:
   - Process the l1 batch number 1
   - Process the l1 batch number 2 and process withdrawals.
   - Reject the l1 batch number 3, and wait for a new valid l1 batch number 3.
8. The Sequencer detects that the verifier network rejected the l1 batch number 3. The Sequencer stops dispatch new l1
   batches data to DA.
9. The Admin execute the rollback and keep the l1 batch number 2 as last valid batch.
10. The Sequencer create a new valid l1 batch number 3.
11. The verifier network verify new l1 batch and ACK that it's valid.
12. The verifier delete the old invalid l1 batch number 3.
13. The verifier network start withdrawal session.
14. Verifier network continue process l1 batches.

#### Case 2: The Admin initial a roll back

1. Start the Sequencer
2. Deposit 10 BTC and wait for l1 batch to be created (`l1 batch 1`).
3. Withdraw 1 BTC and wait for l1 batch to be created (`l1 batch 2`).
4. Withdraw 1 BTC and wait for l1 batch to be created (`l1 batch 3`).
5. Withdraw 1 BTC and wait for l1 batch to be created (`l1 batch 4`).
6. Stop the sequencer
7. Run this sql query to delete the proof commit entry (simulate batch not yet committed)

```sql
UPDATE via_l1_batch_inscription_request SET commit_proof_inscription_id=NULL WHERE l1_batch_number=3
```

8. Run the rollback cmd `make rollback batch=2`, the `**batch**` is the number of the last valid batch we want keep. If
   ENVs are not loaded use this cmd to export the via.env
```sh
   export $(grep -v '^#' etc/env/target/via.env | xargs -d '\n')
```
9. The batch 3 and 4 are deleted and all related data.
10. Restart the Sequencer, a new l1 batch number 3 is created.

## Checklist

- [X] Documentation comments have been added / updated.


core/bin/via_block_reverter/src/main.rs
```diff
@@ -0,0 +1,300 @@
...
+                            .await?,
+                    );
+                }
+            }
+            if rollback_tree {
+                block_reverter.enable_rolling_back_merkle_tree(db_config.merkle_tree.path);
+            }
+            if rollback_sk_cache {
+                block_reverter.add_rocksdb_storage_path_to_rollback(db_config.state_keeper_db_path);
+            }
```
irnb, line 221: maybe we can use builder pattern in here, it's gonna be something looks like this  
```
let block_reverter = ViaBlockReverterBuilder::new(NodeRole::Main, connection_pool)
    .with_postgres_rollback(rollback_postgres)
    .with_tree_rollback(rollback_tree, db_config.merkle_tree.path)
    .build()?;
```

core/node/via_block_reverter/src/lib.rs
```diff
@@ -0,0 +1,457 @@
...
+            .await?;
+        tracing::info!("Rolling back storage logs");
+        transaction
+            .storage_logs_dal()
+            .roll_back_storage_logs(last_l2_block_to_keep)
+            .await?;
+        tracing::info!("Rolling back Ethereum transactions");
+        transaction
+            .eth_sender_dal()
+            .delete_eth_txs(last_l1_batch_to_keep)
```
irnb, line 310: we shouldn't replace this with new logic related to inscription transaction?

via_verifier/node/via_zk_verifier/src/lib.rs
```diff
@@ -330,4 +339,17 @@ impl ViaVerifier {
             Ok(is_valid)
         }
     }
+
+    // This code is triggred only when dev.
+    async fn verification_invalid_l1_batch_numbers(
```
irnb, line 344: Since we’re using a dummy proof on the local machine, where `proof_data.should_verify = false`, this method will be called on the local machine after merging this PR.

Additionally, in the config, we have:

`test_zk_proof_invalid_l1_batch_numbers = [3]`

This means that on the local machine, proof verification of  L1 batch number 3 will always return a false result, which could cause issues. Am I understanding this correctly, or am I missing something?

To fix this, we could introduce a new config option:

`should_test_rollback = false`

It would default to false, and if someone wants to test rollback behavior, they can set it to true.

irnb (approved):
great job man 🔥 😄

core/bin/via_block_reverter/src/main.rs
```diff
@@ -0,0 +1,300 @@
...
+};
+use zksync_core_leftovers::temp_config_store::decode_yaml_repr;
+use zksync_dal::{ConnectionPool, Core};
+use zksync_env_config::{object_store::SnapshotsObjectStoreConfig, FromEnv};
+use zksync_object_store::ObjectStoreFactory;
+use zksync_types::L1BatchNumber;
+
+#[derive(Debug, Parser)]
+#[command(author = "Matter Labs", version, about = "Block revert utility", long_about = None)]
+struct Cli {
```
irnb, line 32: Has the design changed?

I couldn’t find the logic related to the rollback signal that we discussed in this document:

[Via Block Reorgs](https://www.notion.so/txfusion/Via-Block-Reorgs-403058260ff24a18b7a10c9e5cc54646?pvs=4#19a8978a806480399e48c864de0f6397).

via_verifier/node/via_zk_verifier/src/lib.rs
mpavlovic-txfusion: I tested this and it was not a problem because it will return false only once for the configured batch and then it will remove it from the array, so next time that batch will not automatically be treated as invalid.

core/bin/via_block_reverter/src/main.rs
0xatomFusion: Niceee, this is a good proposal to refactor the code, I'll create a task on notion so we can work on this :)

core/node/via_block_reverter/src/lib.rs
0xatomFusion: I removed this part as it's not used by the sequencer [commit](https://github.com/vianetwork/via-core/pull/128/commits/f43f9c8ed8298c97c89fea398170e51f513ed75a)

Makefile
```diff
@@ -188,3 +188,14 @@ clean:
 	@echo "$(YELLOW)Cleaning the project, removing images, volumes, and generated files...$(RESET)"
 	@echo "------------------------------------------------------------------------------------"
 	@$(CLI_TOOL) clean
+
+# Require 'batch' args as input, ex: `make rollback batch=2`
```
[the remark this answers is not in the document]
0xatomFusion: Make sense, fixed in this [commit](https://github.com/vianetwork/via-core/pull/128/commits/6a0159414696d1d6a7cbbf3d25ccd23e35c5f1f2)

PR #131, opened 2025-02-24
fix: do a zk verification before finalize an l1 batch

0xatomFusion:
## What ❔

- Force the verifier to do a zk verification before finalize an l1 batch even if the threshold is reached.
- Add the zk verification threshold to the ENVs.

## Why ❔
Before, the verifier can finalize and l1 batch without doing a zk verification.

## Checklist


etc/env/configs/via_base.toml
```diff
@@ -12,6 +12,8 @@ bootstrap_txids = []
 actor_role = "Sequencer"
 # Number of blocks that we should check when restarting the service.
 btc_blocks_lag = 1000
+# The agreement threshold required for the verifier to finalize an L1 batch.
+zk_agreement_threshold = 1.0
```
irnb, line 16: `let reached_threshold = current_threshold >= threshold;`

based on this logic a faulty (dishonest) verifier could inscribe `NO OK` for  proof validation and force sequencer to revert a valid L1 Batch with Proof 

what do you think about this?

0xatomFusion: This config for devnet and it's possible to update this value when deploy to prod.

irnb: I don’t see the point—if we’re going to switch it back to `0.5` in production, why should we set it to `1.0` in devnet?

0xatomFusion: Done set to 0.5

mpavlovic-txfusion (approved):

irnb (approved):

core/node/via_btc_watch/src/lib.rs
```diff
@@ -54,6 +52,7 @@ impl BtcWatch {
         poll_interval: Duration,
         btc_blocks_lag: u32,
         actor_role: &ActorRole,
+        throshold: f64,
```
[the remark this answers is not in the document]
0xatomFusion: fixed

PR #148, opened 2025-03-17
Feat/via sequencer upgradable

0xatomFusion:
## What ❔

- Implement the **Sequencer** upgrade flow.
- Added new inscription responsible to upgrade the system contracts.
- Added a new upgrade message processor to the BTC watch.
- Added CLI to execute upgrades with doc.
- Fix BTC sender to fetch the last protocol version from the DB.

## Testing
To test the upgrade flow, switch to this [branch](https://github.com/vianetwork/via-core/pull/149) and follow the PR README. 

## Checklist


core/node/via_btc_watch/src/lib.rs
```diff
@@ -128,6 +144,7 @@ impl BtcWatch {
             last_processed_bitcoin_block,
             bridge_address,
             next_expected_priority_id,
+            governance_address: sequencer_address,
```
mpavlovic-txfusion, line 147: Is this just temporary, so we are able to test easier?

0xatomFusion: No, this is not for testing. This is the wallet responsible to create the upgrade inscription. We are using the `Sequencer` wallet as the gov wallet.
I was thinking probably to use a new wallet responsible to create upgrades... it's not mandatory because the upgrade type we are executing is system upgrade, such upgrade is mandatory and should not go through a voting process.
I think to make it clear I will rename it to `sequencer` wallet, wdyt?

0xatomFusion: Looks like this logic is not required as we are already doing a check in the [btc client](https://github.com/vianetwork/via-core/blob/72382fd51ea6d35a090d870a28feb8a8a1984aba/core/lib/via_btc_client/src/indexer/mod.rs#L370-L371), I'll remove it.

mpavlovic-txfusion (approved):

PR #150, opened 2025-03-18
Feat/via verifier upgradable

0xatomFusion:
## What ❔

- Add new protocol version dal to track the Sequencer upgrades.
- Add logic to verify if the verifier is using the same Sequencer protocol version.
- Add message processor to the verifier btc watch.
- Add logic to the ZK verifier to check if a protocol upgrade was executed and switch the verifier to the last protocol version.
- Fix a bug in the coordinator API, a wrong function was used to check if session is in progress.
- Add logic to the coordinator API to check the verifier protocol version during a session.

## Why ❔

When the Sequencer upgrades to a new protocol version, the verifier network should detect and respond accordingly. Verifiers that have not upgraded should be rejected from the network.

## Checklist

- [X] Documentation comments have been added / updated.


via_verifier/lib/verifier_dal/migrations/20250315203029_protocol_version.up.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TABLE IF NOT EXISTS protocol_versions (
```
mpavlovic-txfusion, line 1: Maybe it would be useful to store the `vk_hash`es, at least the `recursion_scheduler_level_vk_hash`, since the verification keys can change with the upgrade. What do you think?

via_verifier/lib/via_verifier_types/src/protocol_version.rs
```diff
@@ -0,0 +1,45 @@
...
+const SEQUENCER_MINOR: ProtocolVersionId = ProtocolVersionId::Version26;
+const SEQUENCER_PATCH: u32 = 0;
+
+/// Bootloader code hash
+const BOOTLOADER_CODE_HASH: &[u8] =
+    &"010008e742608b21bf7eb23c1a9d0602047e3618b464c9b59c0fba3b3d7ab66e".as_bytes();
+
+/// default aa code hash
+const DEFAULT_AA_CODE_HASH: &[u8] =
+    &"01000563374c277a2c1e34659a2a1e87371bb6d852ce142022d497bfb50b9e32".as_bytes();
```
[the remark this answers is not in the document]
0xatomFusion: The goal is to embed these environment variables within the Docker image. Exposing them as environment variables would allow manipulation, potentially leading to a verifier misbehaving due to an incorrect Sequencer version. The decision to hardcode these values is inspired by the Ethereum node, where the node version is embedded directly in the [[binary](https://github.com/ethereum/go-ethereum/blob/master/version/version.go)](https://github.com/ethereum/go-ethereum/blob/master/version/version.go).

via_verifier/lib/verifier_dal/migrations/20250315203029_protocol_version.up.sql
0xatomFusion: Yeah, this make sense to add it

mpavlovic-txfusion (approved):

PR #165, opened 2025-04-14
fix: improve musig2 nonce seed generation

mpavlovic-txfusion:
## What ❔

- Improve MuSig2 nonce generation (first round) security.

## Why ❔

- It is recommended to use cryptographically secured OS level random number generator for generating nonce seed (`OsRng`), so that there is no possibility of reusing the same seed in multiple signing sessions.


0xatomFusion (approved):
lgtm

PR #166, opened 2025-04-14
fix: btc_watch logic index l1 starting block

0xatomFusion:
## What ❔

- Fix the logic where the indexer start indexing the L1 blocks (Sequencer and verifier).
- Adding pagination logic to the L1 blocks selection to avoid memory issues due to selecting many blocks at a time.
- Added a new table `via_indexer_metadata` to track the last indexed L1 block number.
- Added logic to allow restart indexing the L1 chain from a specific L1 block number. This logic can in create duplicates transactions as it's handled by the message processor.

## Why ❔

- At some point we will need to restart the indexer at a specific L1 block number.
- Avoid loading a lots of L1 blocks in memory during the message processing checks.

## Testing
1. Start the sequencer.
2. Deposit BTC.
3. You should see this in DB
Screenshot from 2025-04-14 13-29-47

4. Stop the sequencer and update the `via.env` and update the `VIA_BTC_WATCH_RESTART_INDEXING=true`  
5. Check the DB you should see that the indexer restarted indexing from 1 with a offset of 10 blocks.
6. The already indexed deposit is not re-inserted.
7. Do same for the coordinator.
8. Done

## Checklist

### DIFF
--- a/core/lib/config/src/configs/via_btc_watch.rs
+++ b/core/lib/config/src/configs/via_btc_watch.rs
@@ -65,7 +65,7 @@ impl ViaBtcWatchConfig {
         &self.actor_role
     }
 
-    /// Returns the number of blocks that we should wait before processing the new blocks.
+    /// Returns the starting L1 block number from which indexing begins.
     pub fn start_l1_block_number(&self) -> u32 {
         self.start_l1_block_number
     }
--- a/core/lib/config/src/configs/via_btc_watch.rs
+++ b/core/lib/config/src/configs/via_btc_watch.rs
@@ -8,6 +8,9 @@ pub enum ActorRole {
     Verifier,
 }
 
+/// Total L1 blocks to process at a time.
+pub const L1_BLOCKS_CHUNK: u32 = 10;
+
 /// Configuration for the Bitcoin watch crate.
 #[derive(Debug, Deserialize, Serialize, Clone, PartialEq)]
 pub struct ViaBtcWatchConfig {
@@ -36,8 +39,11 @@ pub struct ViaBtcWatchConfig {
     /// Role of the actor. SEQUENCER or VERIFIER.
     pub actor_role: ActorRole,
 
-    /// Number of blocks that we should wait before processing the new blocks.
-    pub btc_blocks_lag: u32,
+    /// The starting L1 block number from which indexing begins
+    pub start_l1_block_number: u32,
+
+    /// When set to true, the btc_watch starts indexing L1 blocks from the "start_l1_block_number".
+    pub restart_indexing: bool,
 
     /// The agreement threshold required for the verifier to finalize an L1 batch.
     pub zk_agreement_threshold: f64,
@@ -60,8 +66,8 @@ impl ViaBtcWatchConfig {
     }
 
     /// Returns the number of blocks that we should wait before processing the new blocks.
-    pub fn btc_blocks_lag(&self) -> u32 {
-        self.btc_blocks_lag
+    pub fn start_l1_block_number(&self) -> u32 {
+        self.start_l1_block_number
     }
 
     /// Returns the RPC URL of the Bitcoin node.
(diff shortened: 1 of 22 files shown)


core/lib/config/src/configs/via_btc_watch.rs
```diff
@@ -60,8 +66,8 @@ impl ViaBtcWatchConfig {
     }
 
     /// Returns the number of blocks that we should wait before processing the new blocks.
```
mpavlovic-txfusion, line 68: Update/remove description

via_verifier/node/via_btc_watch/src/lib.rs
```diff
@@ -84,31 +87,27 @@ impl VerifierBtcWatch {
...
+            .via_indexer_dal()
+            .get_last_processed_l1_block(VerifierBtcWatch::module_name())
+            .await? as u32;
+
+        if last_processed_bitcoin_block == 0 || restart_indexing {
+            storage
+                .via_indexer_dal()
+                .init_indexer_metadata(VerifierBtcWatch::module_name(), start_l1_block_number)
+                .await?;
+            last_processed_bitcoin_block = start_l1_block_number;
```
mpavlovic-txfusion, line 109: We should probably go with `last_processed_bitcoin_block = start_l1_block_number - 1` ?

core/node/via_btc_watch/src/lib.rs
```diff
@@ -89,31 +95,28 @@ impl BtcWatch {
...
+            .via_indexer_dal()
+            .get_last_processed_l1_block(BtcWatch::module_name())
+            .await? as u32;
+
+        if last_processed_bitcoin_block == 0 || restart_indexing {
+            storage
+                .via_indexer_dal()
+                .init_indexer_metadata(BtcWatch::module_name(), start_l1_block_number)
+                .await?;
+            last_processed_bitcoin_block = start_l1_block_number;
```
mpavlovic-txfusion, line 118: Same here

core/lib/config/src/configs/via_btc_watch.rs
0xatomFusion: Fixed in [commit](https://github.com/vianetwork/via-core/pull/166/commits/026b35e13df3c08362da2edd9db950ea0acf2f11)

via_verifier/node/via_btc_watch/src/lib.rs
0xatomFusion: Fixed in [commit](https://github.com/vianetwork/via-core/pull/166/commits/026b35e13df3c08362da2edd9db950ea0acf2f11)

core/node/via_btc_watch/src/lib.rs
0xatomFusion: Fixed in [commit](https://github.com/vianetwork/via-core/pull/166/commits/026b35e13df3c08362da2edd9db950ea0acf2f11)

mpavlovic-txfusion (commented):
Can you explain and point to the logic that actually makes sure that we don't process the  inscriptions that were already processed and vice versa during indexing restarting?

0xatomFusion:
> Can you explain and point to the logic that actually makes sure that we don't process the inscriptions that were already processed and vice versa during indexing restarting?

The logic to ignore already processed transactions is handled in the message processing logic.

### Sequencer
- [Deposit](https://github.com/vianetwork/via-core/blob/260806a1831a191f35e901773e120320bb6f9f5d/core/node/via_btc_watch/src/message_processors/l1_to_l2.rs#L54-L57)
- [Vote](https://github.com/vianetwork/via-core/blob/260806a1831a191f35e901773e120320bb6f9f5d/core/node/via_btc_watch/src/message_processors/votable.rs#L39-L41)

### Verifier
- [Votable transaction](https://github.com/vianetwork/via-core/blob/260806a1831a191f35e901773e120320bb6f9f5d/via_verifier/node/via_btc_watch/src/message_processors/verifier.rs#L71), the transaction is ignored if already exists in DAL.

- [Deposit](https://github.com/vianetwork/via-core/blob/260806a1831a191f35e901773e120320bb6f9f5d/via_verifier/node/via_btc_watch/src/message_processors/l1_to_l2.rs#L67-L71)
- [Vote](https://github.com/vianetwork/via-core/blob/260806a1831a191f35e901773e120320bb6f9f5d/via_verifier/node/via_btc_watch/src/message_processors/verifier.rs#L114), the logic is handled in DAL.

mpavlovic-txfusion (approved):

PR #168, opened 2025-04-16
Fix/via l1 l2 message processor

0xatomFusion:
## What ❔

- Refactor the l1 to l2 message processor.
- The `l2 gas price` is now hard-coded for L1 deposits. This modification allows to the `Sequencer` and `verifier network` to calculate the correct `L2 canonical transaction hash`. This is important because it's used during the Priority-id verification.
If the l2 gas price is higher than the hard-coded value, the L1TX is still processed as it has priority in the [SQL query](https://github.com/vianetwork/via-core/blob/718fa13f7296cd5d604d1923d9a834d89bd00611/core/lib/dal/src/transactions_dal.rs#L1757-L1761) 
- Fix the bug related to deposit tokens to the system contracts address. Now, all address less than the [MAX_SYSTEM_CONTRACT_ADDRESS](https://github.com/vianetwork/era-contracts/blob/main/system-contracts/contracts/Constants.sol#L29) +1 are rejected, the L1Tx is ignored and no transaction is created.
- Fix the API config. Use Options for via config instead of DataType to be able to init the ZkSync orchestation layer.

## Why ❔

- When using the l2 gas price from the API for deposits it can leads to wrong `L2 canonical transaction hash`, this can cause the verifier network to reject all the blocks.

## Testing
- Start the Sequencer and coordinator
- Try deposit to address between `0x00000....0000000000` and `0x00000....0000010000`, the `via` cmd will fail during the simulation.
- Update the via cli code and comment this [code](https://github.com/vianetwork/via-core/blob/718fa13f7296cd5d604d1923d9a834d89bd00611/infrastructure/via/src/token.ts#L135-L145) and re^place by `const gasCost = BigInt(300_000)`
- Build the via cli and try again. You should see such `WARN` on terminal
Screenshot from 2025-04-16 15-42-46

- Check the database to make sure no transaction was inserted in Sequencer and coordinator db.
- Now, try deposit to any address bigger than  `0x00000....0000010000`, the transaction should pass.
- The verifier should verify and finalize the l1 batch.
- Done :)

## Checklist

### DIFF
--- a/core/lib/types/src/l1/via_l1.rs
+++ b/core/lib/types/src/l1/via_l1.rs
@@ -1,0 +1,99 @@
+use std::str::FromStr;
+
+use zksync_basic_types::{Address, PriorityOpId, H160, U256};
+use zksync_system_constants::REQUIRED_L1_TO_L2_GAS_PER_PUBDATA_BYTE;
+use zksync_utils::address_to_u256;
+
+use super::{L1Tx, L1TxCommonData, OpProcessingType, PriorityQueueType};
+use crate::{
+    abi::L2CanonicalTransaction, helpers::unix_timestamp_ms, Execute, PRIORITY_OPERATION_L2_TX_TYPE,
+};
+
+/// Eth 18 decimals - BTC 8 decimals
+const MANTISSA: u64 = 10_000_000_000;
+
+/// Deposit default L2 gas price.
+const MAX_FEE_PER_GAS: u64 = 120_000_000;
+
+/// Gas limit to required to execute a deposit.
+const GAS_LIMIT: u64 = 300_000;
+
+/// The minimum address that can be used as l2 receiver address.
+const MIN_VALID_L2_RECEIVER_ADDRESS: &str = "0x0000000000000000000000000000000000010001";
+
+#[derive(Debug, Clone)]
+
+pub struct ViaL1Deposit {
+    pub l2_receiver_address: Address,
+    pub amount: u64,
+    pub calldata: Vec<u8>,
+    pub serial_id: PriorityOpId,
+    pub l1_block_number: u64,
+}
+impl ViaL1Deposit {
+    pub fn is_valid_deposit(&self) -> bool {
+        self.l2_receiver_address >= H160::from_str(MIN_VALID_L2_RECEIVER_ADDRESS).unwrap()
+    }
+
+    pub fn l1_tx(&self) -> Option<L1Tx> {
+        if !self.is_valid_deposit() {
+            return None;
+        }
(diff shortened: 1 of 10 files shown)


core/lib/types/src/l1/via_l1.rs
```diff
@@ -0,0 +1,99 @@
...
+const MANTISSA: u64 = 10_000_000_000;
+
+/// Deposit default L2 gas price.
+const MAX_FEE_PER_GAS: u64 = 120_000_000;
+
+/// Gas limit to required to execute a deposit.
+const GAS_LIMIT: u64 = 300_000;
+
+/// The minimum address that can be used as l2 receiver address.
+const MIN_VALID_L2_RECEIVER_ADDRESS: &str = "0x0000000000000000000000000000000000010001";
```
mpavlovic-txfusion, line 22: `MAX_SYSTEM_CONTRACT_ADDR=0x000000000000000000000000000000000000ffff` meaning that probably the `MIN_VALID_L2_RECEIVER_ADDRESS` should be `0x0000000000000000000000000000000000010000`.

Also I tested this and for all addresses from kernel space I got this error:
`Error: virtual machine entered unexpected state. please contact developers and provide transaction details that caused this error. Error description: Assertion error: from in kernel space (payload={ "id": 2, "jsonrpc": "2.0", "method": "zks_estimateGasL1ToL2", "params": [ { "data": "0x", "eip712Meta": { "factoryDeps": [  ], "gasPerPubdata": "0x0320" }, "from": "0x000000000000000000000000000000000000f000", "to": "0x000000000000000000000000000000000000f000", "type": "0x71", "value": "0xbebc200" } ] }, error={ "code": 3, "data": "0x", "message": "virtual machine entered unexpected state. please contact developers and provide transaction details that caused this error. Error description: Assertion error: **from in kernel space**" }, code=3, version=6.13.4)`

And for `0x0000000000000000000000000000000000010000` I get:
`Error: execution reverted (payload={ "id": 2, "jsonrpc": "2.0", "method": "zks_estimateGasL1ToL2", "params": [ { "data": "0x", "eip712Meta": { "factoryDeps": [  ], "gasPerPubdata": "0x0320" }, "from": "0x0000000000000000000000000000000000010000", "to": "0x0000000000000000000000000000000000010000", "type": "0x71", "value": "0xbebc200" } ] }, error={ "code": 3, "data": "0x", "message": "**execution reverted**" }, code=3, version=6.13.4)`

Looks to me that bootloader anyway doesn't allow the addresses from the kernel space and that we introduce the bug to rejects the `0x0000000000000000000000000000000000010000` as well?

PR #173, opened 2025-04-27
Fix: add wallet endpoint to the btc rpc url

0xatomFusion:
## What ❔

- Add the wallet endpoint to the bitcoin RPC url. 

## Why ❔

To track and list he utxos, the wallet needs to be added to the node, this will index the utxos. And to list those utxos on  `Testnet` and `Bitcoin` networks we have to set the wallet endpoint `http://127.0.0.1:8332/wallet/testwallet`,
Without this fix, the node will returns empty list of utxos.

## Checklist

### DIFF
--- a/core/lib/env_config/src/via_wallets.rs
+++ b/core/lib/env_config/src/via_wallets.rs
@@ -17,18 +17,26 @@ impl FromEnv for ViaWallets {
     fn from_env() -> anyhow::Result<Self> {
         let wallets = Wallets::from_env()?;
 
+        let btc_sender_address = pk_from_env(
+            "VIA_BTC_SENDER_WALLET_ADDRESS",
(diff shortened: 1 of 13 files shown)


core/lib/env_config/src/via_wallets.rs
```diff
@@ -17,18 +17,26 @@ impl FromEnv for ViaWallets {
     fn from_env() -> anyhow::Result<Self> {
         let wallets = Wallets::from_env()?;
 
+        let btc_sender_address = pk_from_env(
```
mpavlovic-txfusion, line 20: `pk_from_env` name looks a bit misleading since it is used to load the addresses as well, but no need to change it now, let improve in the future.

mpavlovic-txfusion (approved):

PR #174, opened 2025-04-29
fix: use the chain id from genesis

0xatomFusion:
## What ❔

- Inject the NetworkId during the genesis setup

## Why
The issue was with the NetworkID which is hardcoded in the core to 270. 
```
/// Default `ERA_CHAIN_ID`. All hyperchains start with this chain id and later on during their registration
/// an "initial upgrade" transaction overrides it with the correct value.
```


mpavlovic-txfusion (approved):

etc/env/base/contracts.toml
```diff
@@ -26,8 +26,8 @@ RECURSION_NODE_LEVEL_VK_HASH = "0x1186ec268d49f1905f8d9c1e9d39fc33e98c74f91d91a2
 RECURSION_LEAF_LEVEL_VK_HASH = "0x101e08b00193e529145ee09823378ef51a3bc8966504064f1f6ba3f1ba863210"
 RECURSION_CIRCUITS_SET_VKS_HASH = "0x18c1639094f58177409186e8c48d9f577c9410901d2f1d486b3e7d6cf553ae4c"
 GENESIS_TX_HASH = "0xb99ebfea46cbe05a21cd80fe5597d97b204befc52a16303f579c607dc1ac2e2e"
-GENESIS_ROOT = "0x5d1b8d6f250d89b77b66a918e3e0f60f64ca959aee0c5f1031575491fd00bfa6"
-GENESIS_BATCH_COMMITMENT = "0xe81e1a4727269fe1ef3e2f8c3f5cfb9aab7c073722c278331b7e017033c13f8f"
+GENESIS_ROOT = "0x3c957ca37179902a8957817ddf1469e08aef3f3f3a45fd41803d2ef61808f640"
```
mpavlovic-txfusion, line 29: I guess these are the values that correspond to the new `25223` chain id, so we should probably add that chain id change to the `chain.toml` file as well, otherwise the genesis error occurs that the calculated root hash is not correct.

PR #175, opened 2025-04-30
Add fee rate limits

mpavlovic-txfusion:
## What ❔

- Add fee rate limits (min, max) for Bitcoin transactions.

## Why ❔

- To prevent high gas fee for inscriptions and avoid fee rate spike issues (mainly occurring on Testnet).


0xatomFusion (approved):
LGTM!

PR #180, opened 2025-05-16
Feat/via estimation fee fallback

0xatomFusion:
## What ❔

- Fetch the fee rate from external APIs when the RPC node failed to returns a valid value. 
- Refactor the inscriber initialization by adding a `btc_client_ressource`

This PR introduce new ENVs
```txt
VIA_BTC_CLIENT_EXTERNAL_APIS="https://mempool.space/testnet/api/v1/fees/recommended"
VIA_BTC_CLIENT_FEE_STRATEGIES="fastestFee"
VIA_BTC_CLIENT_USE_RPC_FOR_FEE_RATE=true
```
## Why ❔

- This PR aims to solve the issue we have on testnet related to not enough data to estimate fee.

### Test 1
Run the example fee rate to trigger the fallback
- Run `make base` and `cargo run --example fee_rate true`

### Test 2
- Start the sequencer
- Update the `VIA_BTC_CLIENT_USE_RPC_FOR_FEE_RATE=false`, this will disable fetching the fee rate from the RPC and use only the external API.
- The sequencer should create inscriptions.
- Done :)

## Checklist

### DIFF
--- a/core/lib/config/src/configs/via_btc_client.rs
+++ b/core/lib/config/src/configs/via_btc_client.rs
@@ -11,7 +11,7 @@ pub struct ViaBtcClientConfig {
     pub external_apis: Vec<String>,
     /// Fee strategies
     pub fee_strategies: Vec<String>,
-    /// Use external api to get the inscription fee rate.
+    /// Use the RPC node as main fee rate provider.
     pub use_rpc_for_fee_rate: Option<bool>,
 }
 
--- a/core/lib/config/src/configs/via_btc_client.rs
+++ b/core/lib/config/src/configs/via_btc_client.rs
@@ -6,7 +6,13 @@ use serde::{Deserialize, Serialize};
 #[derive(Debug, Deserialize, Serialize, Clone, PartialEq)]
 pub struct ViaBtcClientConfig {
     /// Name of the used Bitcoin network
-    network: String,
+    pub network: String,
+    /// External fee APIs
+    pub external_apis: Vec<String>,
+    /// Fee strategies
+    pub fee_strategies: Vec<String>,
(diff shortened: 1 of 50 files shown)


core/lib/config/src/configs/via_btc_client.rs
```diff
@@ -6,7 +6,13 @@ use serde::{Deserialize, Serialize};
 #[derive(Debug, Deserialize, Serialize, Clone, PartialEq)]
 pub struct ViaBtcClientConfig {
     /// Name of the used Bitcoin network
-    network: String,
+    pub network: String,
+    /// External fee APIs
+    pub external_apis: Vec<String>,
+    /// Fee strategies
+    pub fee_strategies: Vec<String>,
+    /// Use external api to get the inscription fee rate.
```
mpavlovic-txfusion, line 14: The flag and the comment are a bit misleading, since this needs to be `false` in order to use external api, so better would be something like `Use node rpc to get fee rate`

mpavlovic-txfusion: Nor sure if we need this flag at all, since we can always try the node rpc first and then fallback to external api if configured, right?

0xatomFusion: It's not required to have this flag, I used it to test purpose but in pod not needed if we always want use the RPC node as first solution.

0xatomFusion: Fixed [commit](https://github.com/vianetwork/via-core/pull/180/commits/ec53c496c4d64d069cecbc50177c17fb05f3c1be)

mpavlovic-txfusion (approved):

PR #208, opened 2025-07-02
Feat/via charge users fee when withdraw

0xatomFusion:
## What ❔

- Added `FeeStrategy` trait with a withdrawal strategy implementation to calculate the required fee for a user to be included in a transaction and integrate it to the musig2 transaction builder. 

- Skip (ignore) withdrawals where the value is insufficient to cover the per-user transaction fee.

- The verifier will validate both the ignored withdrawals and the fee_rate used to construct the transaction.

- In the case of empty withdrawals, the verifier network will insert an empty hash to mark the session as processed. A transaction is not submitted in these cases to prevent a potential vector attack where a user could spam the network with small withdrawals. Even if the user incurs a loss, the verifier would still be forced to submit a transaction for each session, resulting in unnecessary transaction fee costs.

- Implemented unit tests to cover the different withdrawals scenarios.

- Add session timeout, this allows the verifiers to have time to finalize a session when there is no valid withdrawals to process (value < fee_per_user)

## Why ❔

This PR introduces new logic to charge users a fee when requesting a withdrawal.
This change is intended to prevent users from spamming the verifier network with small withdrawal amounts, which could lead to unnecessary loss of funds.

## Testing
0. Build via CLI
1. Run the sequencer, coordinator and verifier.
2. Stop the sequencer and increase the sequencer sealing time config
```
CHAIN_STATE_KEEPER_BLOCK_COMMIT_DEADLINE_MS=30000
CHAIN_STATE_KEEPER_MINIBLOCK_COMMIT_DEADLINE_MS=5000
```
3. Deposit 100 BTC

### Test-1
1. Withdraw 1 BTC to `bcrt1qx2lk0unukm80qmepjp49hwf9z6xnz0s73k9j56`
2. Withdraw 1 BTC to  `bcrt1q8tuqv885kehnzucdfskuw6mrhxcj7cjs4gfk5z`
**Expected Output:** Each user should receive less than 1 BTC and the total fee is split between both users.

Screenshot from 2025-07-06 11-13-45

### Test-2
1. Withdraw 0.0000005 BTC to `bcrt1qx2lk0unukm80qmepjp49hwf9z6xnz0s73k9j56`
2. Withdraw 1 BTC to  `bcrt1q8tuqv885kehnzucdfskuw6mrhxcj7cjs4gfk5z`
**Expected Output:** user-2 should receive less than 1 BTC and he pays the total fee. User1 receive nothing.

Screenshot from 2025-07-06 10-55-21

### Test-3
1. Withdraw 0.0000005 BTC to `bcrt1qx2lk0unukm80qmepjp49hwf9z6xnz0s73k9j56`
2. Withdraw 0.0000005 BTC to  `bcrt1q8tuqv885kehnzucdfskuw6mrhxcj7cjs4gfk5z`
**Expected Output:** No one receive tokens and no transaction is created. Check the database to see 0x000...0 hash inserted in the coordinator and verifier.

### Test-4
1. Withdraw 1 BTC to `bcrt1qx2lk0unukm80qmepjp49hwf9z6xnz0s73k9j56`
2. Withdraw 1 BTC to  `bcrt1qx2lk0unukm80qmepjp49hwf9z6xnz0s73k9j56`
**Expected Output:** The same user should receive 2 BTC - fee 

## Checklist


via_verifier/lib/via_musig2/src/fee.rs
```diff
@@ -0,0 +1,202 @@
...
+        let input_size = 148_u64 * u64::from(input_count);
+        // approximate size per output +2 (+1 for OP_RETURN, +1 for potential change)
+        let output_size = 34_u64 * u64::from(output_count + 2);
+
+        let total_size = base_size + input_size + output_size;
+        let fee = fee_rate * total_size;
+
+        let remainder = fee % std::cmp::max(output_count, 1) as u64;
+
+        Ok(Amount::from_sat(fee + remainder))
```
mpavlovic-txfusion, line 24: Not sure about the intention here, I guess this is to ensure the fee is evenly divisible by the number of outputs?

But in that case, I think logic needs to be adjusted.
Example:
fee = 1000, output_count = 3
=> remainder = 1 and 1001 still not divisible by 3

via_verifier/lib/via_musig2/src/fee.rs
```diff
@@ -0,0 +1,202 @@
...
+                    value: output.value,
+                });
+                total_value_needed += value
+            }
+
+            if new_outputs.len() != outputs.len() {
+                // Output set has changed, need to re-estimate fee
+                outputs = new_outputs;
+                continue;
+            }
```
mpavlovic-txfusion, line 82: One suggestion, maybe we can optimize this logic a bit and not create `new_outputs` vector every time, but compare `if new_outputs_with_fee.len() != outputs.len()` and then just recreated the vector that we need to pass as the new `outputs` by adding back the `fee_per_user` to the value.

via_verifier/lib/via_musig2/src/fee.rs
0xatomFusion: [Fixed](https://github.com/vianetwork/via-core/pull/208/commits/457037641dbce321fe4c03e7c2c40d0ea761ba39)

mpavlovic-txfusion (approved):

PR #210, opened 2025-07-09
Feat/via deterministic priority

0xatomFusion:
## What ❔

- Adding the transaction index and the output vout to the inscription CommonData. Those information are required to compute the priorityId.
```rust
pub fn priority_id(&self) -> PriorityOpId {
        // 28 bits for block (268M blocks = 5,100 years with block time = 10min)
        // 24 bits for tx_index (16M transactions per block)
        // 12 bits for vout (4,096 outputs per transaction)
        // |_______block_number_______|_______tx_index_______|_______output_vout_______|
        PriorityOpId(
            ((self.l1_block_number as u64 & 0xFFFFFFF) << 36)
                | ((self.tx_index as u64 & 0xFFFFFF) << 12)
                | (self.output_vout as u64 & 0xFFF),
        )
    }
```

- Update the Verifier btc_watch.
- Update the sequencer btc_watch.
- Update the indexer btc_watch.

## Why ❔

Before this PR, the priorityId was retrieved from the database. This approach made the btc_watch component harder to scale and introduced potential edge cases such as the same transaction being inserted twice due to stale or outdated data (pod running with old priorityId).

To address this, the logic for generating priorityId has been refactored to be deterministic. It now derives directly from on-chain data: specifically the block number, transaction index, and output index (vout). By using these immutable blockchain values, we ensure that each priorityId is both unique and monotonically increasing, which eliminates the risk of duplication and improves the scalability of the system.

## Testing
1. Run the sequencer, the coordinator, the verifier and the indexer.
2.  Deposit and withdraw
3. All batches are processed and verified.

## Checklist


core/lib/types/src/l1/via_l1.rs
```diff
@@ -52,6 +53,18 @@ impl ViaL1Deposit {
     fn value(&self) -> U256 {
         U256::from(self.amount) * U256::from(MANTISSA)
     }
+
+    pub fn priority_id(&self) -> PriorityOpId {
+        // 28 bits for block (268M blocks = 5,100 years when block time = 10min)
+        // 24 bits for tx_index (16M transactions per block)
+        // 12 bits for vout (4,096 outputs per transaction)
```
mpavlovic-txfusion, line 60: After researching maybe I would rather go with this:
```
// 20 bits for tx_index (1M transactions per block)
// 16 bits for vout (65k outputs per transaction)
```
The reason is that current TPS for Bitcoin is around 7, meaning that it could support around 600k tx/day. So, even if this is scaled significantly and we have some peak periods, it is really unlikely that it would reach 1M tx within one block (10 minutes). 

On the other hand, more then 4k outputs in one tx is theoretically possible for some complex tx.

What do you think?

0xatomFusion: It make sense to increase the vout bites to 16 as it's possible to overflow with the current setup.
I updated the logic and added unit test and priority Id validation in case of overflow. 
[Commit](https://github.com/vianetwork/via-core/pull/210/commits/940643935f56fe8d82bf4c05cd12fcd6150a424a)

via_indexer/node/indexer/src/message_processors/deposit.rs
```diff
@@ -59,13 +59,12 @@ impl MessageProcessor for L1ToL2MessageProcessor {
                     continue;
                 }
 
-                let serial_id = next_expected_priority_id;
```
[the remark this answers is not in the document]
0xatomFusion: [Fixed](https://github.com/vianetwork/via-core/pull/210/commits/cea6b803d72a1fc8a394671e5fd8495a1d0b32ac)

PR #212, opened 2025-07-17
Feat/via split withdrawal transaction

0xatomFusion:
## What ❔
- Split an l1_batch request into multiple transactions based on the total weight allowed per transaction (controlled by a new ENV variable).
- Refactored the `withdrawal session` logic to support multiple `bridge_tx` entries per `l1_batch`.
- Included the `index` in each `bridge withdrawal` for better tracking and **traceability.**
- Added a new `debug` subcommand to the `VIA CLI` to allow the creation of multiple `deposits` and `withdrawals`.
- Introduced the `via_bridge_tx` table with a `ONE_2_MANY` relationship to the `via_votable_transaction` table. This table stores the bridge transactions.
- Fixed an issue to release the read MUTEX LOCK earlier during session creation, improving concurrency performance.

## Why ❔
Previously, a musig2 transaction could exceed the max_weight_unit limit allowed by the Bitcoin node, potentially causing the network to stop processing withdrawals.
With this update, when the weight limit is reached, withdrawals are automatically split into multiple transactions to stay within the allowed limits.

### Example
```txt
Let's say the current transaction total weight unit is 240000 bytes and the tx_max_weight is set to 100000. In this case the verifier network creates 3 transactions, each one will contains 80000 WU of data
```

Screenshot from 2025-07-19 12-08-55

You can notice in this example that the batch 8 (votable_transaction_id=8) has 3 transactions.
 
## Testing
1. Start the sequencer, coordinator and verifier.
2. Update the via.env and increase the following ENVs (make sure to restart the sequencer)
```sh
CHAIN_STATE_KEEPER_BLOCK_COMMIT_DEADLINE_MS=60000
CHAIN_STATE_KEEPER_MINIBLOCK_COMMIT_DEADLINE_MS=10
```
Update the coordinator and verifier ENV (make sure to restart the services)
```sh
VIA_VERIFIER_MAX_TX_WEIGHT=2500
```
 Restart the services.
 
### Test 1
- Create multiple small deposits
```sh
via debug deposit-many --amount 20 --receiver-l2-address 0x36615Cf349d7F6344891B1e7CA7C72883F5dc049 --count 20
```
This cmd will split the `amount` by the `count`, in this case 20 deposits will be created with 1 BTC per deposit.

- Create multiple via debug withdraw-many --amount 18 --count 100
```sh
via debug withdraw-many --amount 18 --count 10
```
This cmd will split the `amount` by the `count` and withdraw to 10 different accounts, each account will receive 1.8 BTC.
**Expected Output:** The coordinator and verifier creates multiple transactions to process the l1 batch withdrawals.
- Copy the output, later you will use it to review if all users received the tokens.

### Test 2
- Withdraw 1 BTC to bcrt1qx2lk0unukm80qmepjp49hwf9z6xnz0s73k9j56
- Withdraw 1 BTC to bcrt1q8tuqv885kehnzucdfskuw6mrhxcj7cjs4gfk5z
**Expected Output:** Each user should receive less than 1 BTC and the total fee is split between both users.

### Test-3
- Withdraw 0.0000005 BTC to bcrt1qx2lk0unukm80qmepjp49hwf9z6xnz0s73k9j56
- Withdraw 1 BTC to bcrt1q8tuqv885kehnzucdfskuw6mrhxcj7cjs4gfk5z
**Expected Output:** user-2 should receive less than 1 BTC and he pays the total fee. User1 receive nothing.

### Test-4
- Withdraw 0.0000005 BTC to bcrt1qx2lk0unukm80qmepjp49hwf9z6xnz0s73k9j56
- Withdraw 0.0000005 BTC to bcrt1q8tuqv885kehnzucdfskuw6mrhxcj7cjs4gfk5z
**Expected Output:** No one receive tokens and no transaction is created. Check the database to see 0x000...0 hash inserted in the coordinator and verifier.

## Checklist


via_verifier/lib/verifier_dal/migrations/20250714093607_bridge_tx.down.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE IF EXISTS bridge_tx;
```
mpavlovic-txfusion, line 1: `via_bridge_tx`?

mpavlovic-txfusion (approved):

PR #333, opened 2025-12-28
feat(via_btc_client): implement Largest-First UTXO selection

romanornr:
## What ❔

Largest First UTXO selection

## Why ❔

Optimized UTXO selection 
Current Greedy implementation adds every UTXO, accumulates all value and counts all inputs 
Largest First gives more value per input 

Largest First over Branch and Bound prob not gonna work cause Commit TX change output `COMMIT_TX_CHANGE_OUTPUT_INDEX = 0` is required to fund the Reveal TX

The Reveal TX uses this change output as fee paying input. No change or 0 change means the Reveal tx can't be funded 

DEF needs to be peer reviewed 

## Checklist

- [ ] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [ ] Code has been formatted via `zkstack dev fmt` and `zkstack dev lint`.

### DIFF
--- a/core/lib/config/src/configs/via_btc_sender.rs
+++ b/core/lib/config/src/configs/via_btc_sender.rs
@@ -44,6 +44,18 @@ pub struct ViaBtcSenderConfig {
 
     /// Whether unconfirmed reveal-change outputs from the in-memory inscriber context may be reused.
     pub allow_unconfirmed_change_reuse: Option<bool>,
+
+    /// Minimum feerate for normal inscription construction.
+    pub min_feerate_sat_vb: Option<u64>,
+
+    /// Minimum feerate when the sender is already operating on a pending chain.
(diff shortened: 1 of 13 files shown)


romanornr:
PR #333 follow-up note

Open follow-up concern to revisit later:
- `MAX_UTXOS_TO_CONSIDER = 100` can theoretically cause false `Insufficient funds` if the wallet has many UTXOs and the truncated smaller tail is still needed to meet the target.
- This was intentionally left unchanged in the current hardening series because it is a separate tradeoff between performance and completeness.
- Future work options:
  1. iterate all candidate UTXOs but stop as soon as target is met
  2. only apply truncation if the retained set can still satisfy the target
  3. make the cap config-backed and monitor real wallet UTXO distributions before changing defaults

Issue #348, opened 2026-04-01
Execute migration for #347: deprecated Mailbox methods -> Bridgehub/L1Nullifier


cryptocake:
Title: Execute migration for issue #347: replace deprecated Mailbox methods with Bridgehub/L1Nullifier flows

Context
zkSync has deprecated Mailbox.requestL2Transaction and Mailbox.finalizeEthWithdrawal. We need to migrate active code paths and keep compatibility where needed.

Scope
1) Active code migration
- Replace deprecated requestL2Transaction usage in loadnext/via_loadnext paths with Bridgehub request path.
- Resolve TODO(EVM-571) migration points tied to shared bridge migration.

2) Withdrawal finalize flow
- Add dual compatibility in withdrawal parsing:
  - legacy finalizeEthWithdrawal selector path
  - new finalizeDeposit-compatible path for L1Nullifier flow
- Select path by config/version or robust payload detection.

3) Documentation and hygiene
- Update docs referencing deprecated methods in active guides/specs.
- Keep archival upgrade artifacts untouched, but mark as historical where appropriate.

4) Regression guard
- Add CI check to block NEW references to deprecated methods/selectors in non-archival dirs.

Acceptance criteria
- No active runtime/test tooling path depends on deprecated methods.
- Withdrawal parser supports both legacy and new flow.
- Integration smoke passes for Bridgehub L1->L2 and withdrawal finalization path.
- CI guard prevents reintroduction.

Proposed PR split
- PR A: loadnext/via_loadnext request path migration
- PR B: withdrawal parser dual compatibility + tests
- PR C: docs + CI deprecation guard

PR template for this work
- Summary
- What changed
- Migration notes (legacy compatibility kept/removed)
- Test plan
  - [ ] Unit tests updated/added
  - [ ] Integration smoke run
  - [ ] No new deprecated refs (CI check)
- Risks and rollback
- Linked issue(s)
  - Closes #347

PR #349, opened 2026-04-01
fix(loadnext): use requestL2TransactionBridgehub in ETH deposit path

cryptocake:
## Summary
Migrate the loadnext ETH deposit request path away from deprecated `requestL2Transaction` to `requestL2TransactionBridgehub` in both loadtest SDK variants.

## Changes
- `core/tests/loadnext/src/sdk/ethereum/mod.rs`
  - switched encoded function from `requestL2Transaction` -> `requestL2TransactionBridgehub`
  - updated EVM-571 comment wording to remove deprecated-method reference
- `core/tests/via_loadnext/src/sdk/ethereum/mod.rs`
  - switched encoded function from `requestL2Transaction` -> `requestL2TransactionBridgehub`
  - updated EVM-571 comment wording to remove deprecated-method reference

## Why
This keeps the loadnext request path aligned with the Bridgehub-compatible method exposed in current ABI surfaces and reduces continued reliance on deprecated Mailbox method naming.

## Validation
- Verified both loadnext SDK files now encode `requestL2TransactionBridgehub`
- Verified no `requestL2Transaction` function string remains in those two files

## Linked issues
- Part of #347
- Ref #348

### DIFF
--- a/core/tests/loadnext/src/sdk/ethereum/mod.rs
+++ b/core/tests/loadnext/src/sdk/ethereum/mod.rs
@@ -407,18 +407,20 @@ impl<S: EthereumSigner> EthereumProvider<S> {
             .base_cost(gas_limit, L1_TO_L2_GAS_PER_PUBDATA, Some(gas_price))
             .await
             .map_err(|e| ClientError::NetworkError(e.to_string()))?;
-        let value = base_cost + operator_tip + l2_value;
+        let mint_value = base_cost + operator_tip + l2_value;
         let tx_data = self.client().encode_tx_data(
-            "requestL2Transaction",
-            (
+            "bridgehubRequestL2Transaction",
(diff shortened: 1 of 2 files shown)


cryptocake:
Good catch on the ABI mismatch. Updated this PR to match the hyperchain ABI shape.

Changes made:
- switched method to `bridgehubRequestL2Transaction`
- switched args to a single `BridgehubL2TransactionRequest` tuple
- included `sender` and `mintValue` in the tuple
- removed `msg.value` from tx options to match the nonpayable signature

Patch is pushed on the same branch.

cryptocake:
Closing this PR due to migration-scope and ABI-semantics ambiguity. Will replace with a stricter, minimal, ABI-verified change.

PR #378, opened 2026-06-12
docs(refork): Phase 0 inventory and v29 seam analysis

romanornr:
# What ❔

Phase 0 deliverables for the refork of zksync-era at `core-v29.20.0`:

- `etc/refork/inventory.csv` + `build_inventory.py` — all 3,457 diverged paths vs fork
  point `f37b84ac75`, categorized (via crates / modified upstream / deleted /
  migrations) and joined against the pin for their v29 fate, including byte-identical
  backport detection.
- `docs/refork/port-inventory.md` — headline findings and the dependency-wave porting
  order for the 39 via crates.
- `docs/refork/seams/01–07` — per-seam mapping of what via patched in v24 to the
  verified v29 extension point.
- `docs/refork/smithers-workflow.md` + `differential-evals.md` — orchestration and
  parity-eval specs for the port execution phase.

## Why ❔

via-core is ~21 months behind upstream; merging is no longer feasible. These artifacts
are the inventory, architecture mapping, and orchestration specs the refork executes
against. Key findings: the refork is additive (via never deleted the ETH stack), 591
diverged paths are pure backports of upstream work, and the forked via crates carry
small deltas that should be re-applied onto v29 crates rather than carried as forks.

## Checklist

- [x] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [x] Documentation comments have been added / updated.

---

Reproduce the inventory:

```bash
git remote add upstream https://github.com/matter-labs/zksync-era.git
git fetch upstream tag core-v29.20.0 --no-tags
python3 etc/refork/build_inventory.py
```

https://claude.ai/code/session_01XoczhkbwY36DmhAwnkN5i4


romanornr:
Review findings addressed in b8ee0fad (all verified empirically before fixing — counts below are from the regenerated `inventory.csv` against the new pin `core-v29.20.0`):

**Rename LOC parsing** (@gemini-code-assist high-priority, also Copilot/CodeRabbit/Codex) — confirmed: `--numstat -M` emits brace-compressed paths and the old handling was a no-op; 560 of 566 R* rows carried `0/0`. Fixed with a `resolve_renamed()` helper. Effect: B/feature 504→583 (473 real after backports), B/wiring 916→837.

**Migration backports misclassified** (Codex) — confirmed, the largest content error: of 71 `via-migration` rows, 32 were cherry-picked upstream migrations (29 verbatim paths in v29, 3 more content-identical). Added migrations now get the same v29 fate check as other additions; D is now 39 via-owned files, backports are classed `A/backported-migration` (zero port work). Seam 06 renumbering no longer queues phantom migrations.

**`moved?` determinism + precision** (Copilot/CodeRabbit/Codex) — fixed beyond the sorted pick: move detection is now content-exact (pre-fork blob hash found at another v29 path → `moved:`), with basename matching demoted to a hint emitted only when the basename is unique in v29 (`moved?`); ambiguous matches stay `gone`. Fates are now: 835 gone / 889 present / 622 identical / 72 moved / 188 moved?. CSV verified byte-identical across runs with different `PYTHONHASHSEED`.

**Self-inclusion of the Phase 0 artifacts** (Codex) — fixed via `EXCLUDE_PREFIXES = ("docs/refork/", "etc/refork/")`; the documented reproduction now returns the checked-in 3,457 rows after the commit lands.

**Nits** (Copilot) — unused `sys` removed; `DictWriter` uses an explicit fieldname list.

Docs citing the affected numbers (`port-inventory.md`, `README.md`, seams 03/06) were reconciled in the same commit, and the pin was bumped `core-v29.19.2` → `core-v29.20.0` (newest core tag; a fate-diff between the two showed zero classification changes, documented in the README's "Pin choice" section).

romanornr:
Second review round addressed in 46e7ea6d — all eight findings verified against the data first, and all eight were valid:

**`fate()` ignored the post-rename path** (CodeRabbit, and Codex's self-referential `moved?` rows are the same root cause) — confirmed and larger than reported: **208 R\* rows** whose new path exists in v29 were misfated `gone`/`moved?` (96 of them self-referentially pointing at their own path). Fixed by checking `new_path in v29_tree` before old-path lineage. Fates now: 726 gone / 1,097 present / 622 identical / 69 moved / 92 moved?.

**Migration backport false positive** (Codex) — confirmed: `witdrawals.down.sql`'s one-line placeholder blob-matched an unrelated upstream down-migration. Backport detection is now path-level only; the pair is back in the via queue (D = 40, backported = 31).

**via-crate over-matching** (Codex) — confirmed: 74 tooling rows (via-playground, `.github` lint, docker dirs, protocol-upgrade scripts) sat in the via-crate bucket. A crate now requires a `Cargo.toml` with a package name under its directory; via-crate = 403 rows across **exactly 39 distinct units** — the headline count is now true by construction.

**`verifier_dal` vs `via_verifier_dal` unit-key mismatch** (CodeRabbit) — fixed as a by-product of the above: unit keys are now the Cargo package names, so every wave name in `port-inventory.md` resolves to a CSV unit (`via_verifier_dal`, `via_verifier_btc_sender`, …).

**CRLF in the checked-in CSV** (Codex) — confirmed (all 3,458 lines); writer now uses `lineterminator="\n"` and `git diff --check` passes on the new commit.

**`config` missing from the wave queue** (Codex) — added as Wave 0 in `port-inventory.md`: via's smart-config structs/secrets land before any crate.

**`SettlementLayer::L1(...)` placeholder** (CodeRabbit) — removed from seam 02; the audit is now binary per consumer: omit the layer, or a via-specific adapter with an explicit invariant that the ETH path stays inert.

Docs citing affected counts (`port-inventory.md` table, README, seam 06) were reconciled in the same commit. CSV remains 3,457 rows and byte-identical across runs with different `PYTHONHASHSEED`.

PR #387, opened 2026-09-07
fix(via-btc): synchronize external-node settlement details

romanornr:
## Why

Via main nodes own Bitcoin inscription and verifier-verdict records. External nodes do not reproduce those records: `BatchStatusUpdater` mirrors the main node's lifecycle metadata into the existing ETH-shaped settlement tables. The active detail handlers nevertheless read the main-node Bitcoin representation for both roles, so an external node could retain commit/proof metadata locally while its block, batch and transaction detail APIs omitted it.

### Background: how a Via batch becomes "settled", and how an external node learns about it

Via is a ZKsync Era fork that settles on Bitcoin instead of Ethereum. The main node seals L2 blocks into L1 batches and then walks each batch through three steps:

1. **Commit.** The batch commitment is inscribed on Bitcoin. The reveal transaction id becomes the batch's commit hash, and the Bitcoin confirmation time becomes `committedAt`.
2. **Prove.** The validity proof is inscribed the same way; its reveal transaction id becomes the prove hash.
3. **Verdict.** A verifier network votes on the proof. The main node stores the outcome in `via_l1_batch_inscription_request.is_finalized`: `NULL` while votes are pending, `TRUE` when accepted, `FALSE` when rejected. A rejected batch is reverted and re-produced by the chain.

The inherited ZKsync code, its database schema and its JSON-RPC API all speak Ethereum: `eth_txs`, `eth_txs_history`, `l1_batches.eth_commit_tx_id / eth_prove_tx_id / eth_execute_tx_id`, and detail responses with `commitTxHash`, `proveTxHash`, `executeTxHash`. Via maps its Bitcoin lifecycle onto those names so that ZKsync tooling, including block explorers, keeps working. Commit and prove map cleanly to the two reveal transaction ids. There is no Bitcoin "execute" transaction, so Via emits a constant placeholder, `0x1111…1111`, as the execute hash once the verifier network has accepted the batch. Explorers treat any non-zero execute hash as "Executed on L1".

An external node is a read replica. It cannot see inscriptions or verifier votes, so every few seconds `BatchStatusUpdater` asks its main node `zks_getL1BatchDetails` for the next batches and copies the hashes and timestamps it receives into the Ethereum-shaped tables (`insert_bogus_confirmed_eth_tx`). That mirror is what the EN's `eth_getBlockByNumber("finalized")`, its health cursors and its Via consistency checker already relied on.

The bug was that the EN's `zks_getL1BatchDetails`, `zks_getBlockDetails` and `zks_getTransactionDetails` handlers read the Via inscription tables, which only a main node fills. On an EN those tables are empty, so the EN answered "sealed, no hashes" for every batch, forever, while its own health said committed, proven and executed. Anything indexing settlement from an EN, such as an explorer, never saw a batch settle.

Fixing the reader exposed two more defects in the same path:

- The mirror deduplicates transactions by hash, and every accepted batch carries the same `0x1111…1111` execute hash. The writer therefore reused the first batch's history row and its `confirmed_at` for every later batch, so a plain reader switch would have shown batch 1's execution time on batch 263.
- The main node reported `verified` plus the placeholder hash whenever `is_finalized` was non-NULL. A rejected batch looked exactly like an accepted one on the wire, and an EN would have mirrored the rejection as an execution.

The fix therefore has four parts: read from the mirror on external nodes and from the inscription tables on main nodes; store each batch's own execution time next to the link instead of on the shared history row; carry the verdict explicitly as `viaIsFinalized` so an EN records an execution only on `true`; and report a missing or negative verdict through health instead of either crashing or staying silent.

```mermaid
flowchart LR
  M["Main node<br/>via_* inscription tables<br/>verifier verdict"]
  U["EN BatchStatusUpdater"]
  T[("EN mirror<br/>l1_batches.eth_*_tx_id<br/>eth_txs / eth_txs_history<br/>via_en_executed_at")]
  A["EN zks_getL1BatchDetails<br/>zks_getBlockDetails<br/>zks_getTransactionDetails"]
  M -- "zks_getL1BatchDetails<br/>+ viaIsFinalized" --> U
  U -- "commit / prove links" --> T
  U -- "execute link + per-batch time<br/>only when viaIsFinalized = true" --> T
  T -- "this PR" --> A
  M -. "before: EN APIs read the EN's empty via_* tables" .-> A
```

One batch through the lifecycle, as the three detail APIs report it:

| Event on the main node | Main-node API | External node before | External node after this PR |
| --- | --- | --- | --- |
| Batch sealed, nothing inscribed | `sealed`, no hashes, execute hash `0x00…00` | `sealed`, no hashes | `sealed`, no hashes |
| Commit inscription confirmed | `commitTxHash` = reveal txid (display byte order), `committedAt` = Bitcoin confirmation time | `sealed`, no hashes | same hash and time as main |
| Proof inscription confirmed | plus `proveTxHash`, `provenAt` | `sealed`, no hashes | same as main |
| Verifier network accepts (`is_finalized = TRUE`) | `verified`, execute hash `0x11…11`, `executedAt` = verdict time, `viaIsFinalized: true` | `sealed`, no hashes | `verified`, `0x11…11`, this batch's own `executedAt`, `viaIsFinalized: true` |
| Verifier network rejects (`is_finalized = FALSE`) | before: `verified` + `0x11…11` (defect); now: `sealed`, `0x00…00`, `viaIsFinalized: false` | `sealed` | `sealed`; updater health `affected` naming the batch until the chain replaces it |
| Batch synced before this PR (legacy link, no per-batch time) | unchanged | `sealed`, no hashes | commit and prove visible; execution withheld until the separate recovery step |

Execution needs a separate acceptance rule. Via reuses `0x1111…1111` as a synthetic execution hash, not a Bitcoin execution transaction. A present but false verifier verdict must not mean successful execution, and a hash-deduplicated history row cannot represent distinct execution times for multiple batches. This change makes the detail APIs use the role-owned local representation and requires affirmative acceptance plus a batch-local timestamp before exposing a newly synchronized Via execution.

These source and synthetic-test results are not live settlement or explorer-recovery proof. In particular, this PR does not retroactively repair block 1037 / batch 263, the main-node finalized-block tag, or an explorer's shared-hash timestamp representation.

## What changed

- The existing `via_zks::ZksNamespace` selects the synchronized local DAL for external-node block, batch and transaction details through `InternalApiConfig.use_synced_settlement`. Main-node readers stay Bitcoin-backed. Both EN config conversions set the internal selector; main-node construction leaves it false. Submission, transaction sinks and proxy behavior are unchanged.
- `L1BatchDetails` carries additive, optional `viaIsFinalized`. Main-node conversions require `Some(true)` rather than mere verdict presence. Unknown verdicts remain omitted; rejected verdicts do not report successful execution.
- `BatchStatusUpdater` withholds unsupported synthetic execution without terminating the node or stopping commit/proof synchronization. Execution remains sequential. Affected health names the blocking batch, for an explicit rejection (`viaIsFinalized: false` with a zero execution hash) as well as for a missing verdict; identical cursor/blocking states do not serialize or republish health. Missing accepted timestamps and storage failures remain errors. A successful empty/missing-batch response can clear affected health without advancing execution; health is diagnostic, not finality proof.
- The nullable `l1_batches.via_en_executed_at` column records the accepted batch's execution time in the existing mirror-writer transaction. Generic execution-link assignment clears any old timestamp. Detail reads withhold synthetic execution when this timestamp is NULL and never borrow another batch's shared history timestamp. The execution chain-id join is keyed off the already-filtered history row, so a withheld execution exposes no execution metadata at all.
- `EthSenderDal::delete_eth_txs` no longer deletes an `eth_txs` row that a retained batch still references. On an external node every accepted execution shares one sentinel row, and the EN runs the upstream block reverter with executed-batch rollback allowed, so a revert past an executed batch previously removed that shared row and the `ON DELETE SET NULL` constraint cleared the execute link on every retained batch. The replacement query is shorter than the three-way `UNION` it replaces.
- Six checked queries have regenerated SQLx metadata against the migrated disposable PostgreSQL schema: the three detail SELECTs, the Execute branch of `set_eth_tx_id`, the writer's timestamp UPDATE and `delete_eth_txs`; two obsolete entries were removed. Test status assertions use enum-pattern matching because the existing status enums do not implement `PartialEq`; production types were not changed to accommodate tests.
- Three unused private upstream detail-handler copies were removed. Regression coverage exercises main-node HTTP → production updater → external-node HTTP, status transitions, rejection with the real main-node response shape, shared-hash timestamps, legacy links, chain-id suppression, shared-row preservation across `delete_eth_txs`, and snapshot recovery. The EN README documents the operating boundary.

## Reuse & Duplication

- Closest existing implementation considered: `via_zks::ZksNamespace::{get_block_details_impl,get_l1_batch_details_impl,get_transaction_details_impl}`, `BatchStatusUpdater`, `EthSenderDal::{insert_bogus_confirmed_eth_tx,delete_eth_txs}`, and `BlocksDal::set_eth_tx_id`.
- Sibling paths inspected: `core/node/api_server/src/web3/namespaces/{via_zks,zks}.rs`; `core/lib/dal/src/{blocks_web3_dal,via_blocks_web3_dal,transactions_web3_dal,via_transactions_web3_dal}.rs`; both `core/bin/{external_node,via_external_node}/src/config/mod.rs`; `core/node/via_btc_watch/src/message_processors/votable.rs`; `via_verifier/node/via_btc_watch/src/message_processors/verifier.rs`; and the two reverters, `core/node/block_reverter/src/lib.rs` (wired by `via_external_node` with `allow_rolling_back_executed_batches`) and `core/node/via_block_reverter/src/lib.rs` (used only by the `via_block_reverter` CLI, which does not call `delete_eth_txs`).
- Search command run during implementation review: `rg -n 'set_eth_tx_id\(|insert_bogus_confirmed_eth_tx\(|delete_eth_txs\(|eth_execute_tx_id\s*=' core via_verifier --glob '*.rs' --glob '*.sql'`; active handler references were also inspected with `rg -n 'get_block_details_impl|get_transaction_details_impl|get_l1_batch_details_impl' core/node/api_server/src/web3`. No GitNexus was used.
- Shared extraction decision: extend the existing updater, writer, setter, cleanup query and active namespace rather than add a poller or proxy. Main-node Bitcoin records and EN display-ordered lifecycle mirrors have different ownership and conversion rules, so their DAL models remain separate. The core watcher writes votes only after sender-owned proof inscriptions exist; the verifier watcher indexes verifier-specific votable records. Neither owns EN mirroring or these detail APIs, so neither watcher is modified. No mapped sibling pair or structural-rule source is changed.
- The repeated SQL sentinel conditions preserve single-statement reads with different block/batch/transaction projections. The removed private upstream copies had no active callers; future upstream merges touching them may require deliberate conflict resolution rather than restoring duplicate handlers.

## Performance, Complexity, and Resource Impact

| Dimension | Result |
| --- | --- |
| Detail reads | Existing lifecycle SELECT and conversions; no additional RPC, SELECT, cache or heap collection. Existing pruning checks and transaction readonly/fallback behavior remain. |
| Accepted synthetic execution | One additional primary-key UPDATE inside the existing transaction/savepoint. The shared history row stays hash-deduplicated. |
| Polling | Existing linear transition consumption and range skipping. A fully proven range blocked at execution probes its boundary rather than scanning every later proven batch. Commit/proof catch-up still proceeds in order. |
| Health | Fixed-size cursor/blocked-batch state. JSON construction and publication occur only when that state changes, not on identical idle/blocked polls. No new worker, mutex or cache. |
| Rollback cleanup | `delete_eth_txs` now scans retained `l1_batches` rows once per revert to exclude ids they still reference (three ids per row through a hash `EXCEPT`). Reverts are rare operator or reorg events; no per-request cost. |
| Storage / migration | Nullable timestamp, no default, index or backfill. Populated timestamps add roughly eight payload bytes per batch plus PostgreSQL tuple/null overhead. DDL still requires an approved lock/rollout window. |
| Audit cost | Approximately +40 nonblank, noncomment production Rust/embedded-SQL lines against the base, excluding test files, documentation, generated SQLx and migration files. That figure includes the removal of about 50 lines of unused upstream handlers; the active fix itself adds roughly 90 lines. |

## Boundaries and non-goals

This is a source/runtime change, not a deployment or historical-recovery operation. No live database, service, wallet, credential, infrastructure or deployment desired state was accessed or changed for validation/publication. Only dedicated synthetic local PostgreSQL databases were migrated. No changes were made to the original source checkout, other candidates, `kube-state`, `helm-charts` or infrastructure configuration.

Legacy execution links with NULL batch-local timestamps remain inconclusive to detail readers, and startup cursors do not revisit them. The ETH-backed finalized tag/startup cursor can still count such links while detail APIs withhold execution. No cursor rewind, backfill, fabricated inscription record or shared-history rewrite is included. ENs do not become a negative-verdict ledger: missing/false upstream verdicts are withheld rather than separately persisted.

The `delete_eth_txs` change also applies to the upstream main-node reverter path: an aggregated transaction whose batches straddle a revert boundary is now kept instead of deleted. Via main nodes do not run `eth_sender`, so this only affects the excluded upstream binary; keeping a mined transaction row that retained batches reference is the safer semantics in either case.

## How to review

1. Follow the internal role selector through both EN config conversions and the three active detail handlers. Confirm that main-node Bitcoin reads, byte-order conversions, pruning and transaction submission/fallback remain separate from EN mirror reads.
2. Follow `viaIsFinalized == Some(true)` from the main-node models through `UpdaterCursor::update_stage`. The sequential batch check comes first; an explicit rejection or a sentinel without acceptance sets the blocked batch and returns; unsupported execution must preserve commit/proof liveness without allowing later batches to skip the execution boundary.
3. Review the migration, execution setter and mirror writer together. The batch-local timestamp and execution link must commit together; ordinary link assignment must invalidate old acceptance. In SQLx metadata, the two CASE-derived `executed_at?` columns have unknown inferred nullability but retain explicit nullable aliases; column types and query parameters are unchanged for the three detail SELECTs.
4. Review `delete_eth_txs` together with `core/node/block_reverter/src/lib.rs` (the reverter the EN runs) and the `ON DELETE SET NULL` constraints on `l1_batches.eth_*_tx_id`. The `via_block_reverter` CLI does not call this cleanup.
5. Read `via_settlement_status_sync_to_all_detail_apis`, `unsupported_via_execution_preserves_progress`, `via_execution_still_requires_timestamp`, and both snapshot-recovery cases. Two accepted batches deliberately share one synthetic hash and retain different execution times; the rejected case uses the real main-node shape (false verdict, zero hash). HTTP was exercised; separate WS runtime parity and live explorer recovery are not claimed.
6. Treat the migration/rollback and historical-recovery decisions below as rollout prerequisites, not permission to execute them.

## Checks run

Local validation used Rust `nightly-2024-08-01`, disposable PostgreSQL 14.23 databases, an isolated Cargo target directory, and no live database URLs. SQLx generation/check used the migrated compile database; test clones used a separate fully migrated template.

### Initial validation on `ba2086dc3135851fc8821db8717dfbe93ee41633`

```sh
# From core/lib/dal, with DATABASE_URL pointing only to the disposable migrated DB:
cargo sqlx prepare --database-url "$DATABASE_URL" -- --tests
cargo sqlx prepare --check --database-url "$DATABASE_URL" -- --tests

# From the candidate root, SQLX_OFFLINE=true:
cargo check --tests -p zksync_dal -p zksync_node_sync -p zksync_node_api_server -p via_external_node -p zksync_snapshots_applier -p zksync_eth_sender
cargo test -p zksync_node_sync --lib batch_status_updater -- --test-threads=1
cargo test -p zksync_node_api_server --lib via_settlement_status_sync_to_all_detail_apis -- --test-threads=1
cargo test -p zksync_node_api_server --lib methods_with_snapshot_recovery -- --test-threads=1

just via-check
git diff --check
gitleaks protect --staged --redact --no-banner
gitleaks git --log-opts="d26426894b52ca7516273b690c58410b1e73aa1a..ba2086dc3135851fc8821db8717dfbe93ee41633" --redact --no-banner
```

- Focused tests: **11 updater + 1 HTTP settlement + 2 snapshot = 14 passed, 0 failed**. Affected test targets compile.
- Focused manual review found no blocker in the regenerated SQLx entries or corrected status assertions. CodeRabbit CLI 0.7.6 completed directory-scoped light reviews of the changed SQLx metadata and API tests with **0 findings** in each. Staged and committed-range secret scans passed.
- Migration down/up rehearsal: the new column count was 0 after down and 1 after up in a disposable schema database. No live rollback was rehearsed.
- Candidate formatting: `rustfmt --edition 2021 --check --config skip_children=true` over the candidate Rust files and Prettier/markdownlint on the changed EN README passed. Whole-repository `zkstack dev fmt` was attempted; unrelated formatter changes were discarded. Whole-repository formatting is not claimed clean.
- The workstation's GCC 16.2.1 fails building bundled RocksDB 8.1.1 without `<cstdint>`. Compilation/tests passed with isolated `CXXFLAGS='-include cstdint'`; no vendored source or build configuration was changed. System-contract artifacts were built with Node 20 and disposable locked JS tooling, outside committed source changes.
- `zkstack dev lint --check` target results: `sol`, `js`, `autocompletion` and `rust-toolchain` passed. Structural lint passed in advisory mode with existing duplicate-export warnings. `via-check-strict` is not required by the changed paths.

### Re-validation on `be6a0a4e506afa6d00b29cfd472ab76367647b93` (review fixes, amended and force-pushed with an exact lease)

The amended commit differs from `ba2086dc` only in the review fixes: the chain-id join, `delete_eth_txs`, the rejection branch and message in `update_stage`, the extended tests, and three regenerated SQLx entries.

```sh
# From core/lib/dal, disposable compile DB:
cargo sqlx prepare --database-url "$DATABASE_URL" -- --tests
cargo sqlx prepare --check --database-url "$DATABASE_URL" -- --tests

# From the candidate root:
cargo test -p zksync_node_sync --lib batch_status_updater -- --test-threads=1
cargo test -p zksync_node_api_server --lib via_settlement_status_sync_to_all_detail_apis -- --test-threads=1
rustfmt --edition 2021 --check --config skip_children=true <the five changed Rust files>
just via-check
git diff --cached --check
gitleaks protect --staged --redact --no-banner
```

- **11 updater + 1 HTTP settlement = 12 passed, 0 failed.** These runs compile `zksync_dal`, `zksync_node_sync` and `zksync_node_api_server` with the new SQLx metadata. SQLx check, formatting of the changed files, structural lint (advisory warnings unchanged), whitespace and secret scans passed; the commit signature verifies.
- Not rerun after the review fixes: the two snapshot-recovery cases, the broader `cargo check --tests` crate set, the migration down/up rehearsal and the lint targets listed above. The changed detail-SELECT text since `ba2086dc` is limited to the execution chain-id join.

**Outstanding baseline lint failures, not waived or suppressed:**

| Gate | Observed failure | Disposition |
| --- | --- | --- |
| Rust lint / root Clippy | `core/lib/types/src/via_wallet.rs:139`: `clippy::clone_on_copy` on `Txid` | Unchanged file; root failure prevented subsequent prover/zkstack Rust lint workspaces from running. |
| TypeScript lint | Seven unused-variable errors in `infrastructure/via/src/init.ts` and `infrastructure/via/src/up.ts` | Unchanged files; compatible locked TypeScript 4.9.5 was installed in disposable tooling before the final run. |
| Contracts lint | Existing `L1SharedBridge.sol` gas-custom-error/style findings | Unchanged contracts gitlink/source. |
| Repository Markdown lint | Existing guide violations, including `docs/via_guides/gov/bridge-upgrade/2.upgrade.md` MD029 | Unchanged guides; the candidate README passes its targeted check. |

### GitHub results

On `ba2086dc` all seven workflows passed ([CI](https://github.com/vianetwork/via-core/actions/runs/34135815117), [Code Quality](https://github.com/vianetwork/via-core/actions/runs/34135811428), [PR Template Check](https://github.com/vianetwork/via-core/actions/runs/34135814932), [PR title](https://github.com/vianetwork/via-core/actions/runs/34135814949), [Via Structural Rules](https://github.com/vianetwork/via-core/actions/runs/34135814917), [Zizmor](https://github.com/vianetwork/via-core/actions/runs/34135814870), [Copilot review workflow](https://github.com/vianetwork/via-core/actions/runs/34135822043)).

On `be6a0a4e5`, [PR Template Check](https://github.com/vianetwork/via-core/actions/runs/34145397809), [PR title](https://github.com/vianetwork/via-core/actions/runs/34145397832), [Via Structural Rules](https://github.com/vianetwork/via-core/actions/runs/34145397797) and [Zizmor](https://github.com/vianetwork/via-core/actions/runs/34145397835) passed; [CI](https://github.com/vianetwork/via-core/actions/runs/34145398025) had finished contract preparation and was still building the four core images when this description was last updated. The Copilot review workflow had not re-run on the amended commit at that time.

The separate [Kilo Code Review check](https://app.kilo.ai/code-reviews/a97ba39b-ebdc-445a-9caa-b89df674625c) ends `ACTION_REQUIRED`: **“Insufficient credits to run review.”** Its review never started and produced no annotations. This is an account/service prerequisite, not a candidate build or test failure; no credits were purchased, rerun requested, or gate disabled.

The active CI builds core images; its broad Rust lint/test workflows are disabled. Successful Actions runs therefore do not erase the local lint failures above or replace the focused test results. No merge readiness or approval is claimed.

### Automated review threads on `ba2086dc` and their disposition in `be6a0a4e5`

| Thread | Verdict | Disposition |
| --- | --- | --- |
| CodeRabbit, `blocks_web3_dal.rs:724`, execution chain id leaks when the sentinel is withheld | Valid, cosmetic on Via (the EN never stores a chain id) | Fixed: both `execute_tx_data` joins key off the filtered `execute_tx` history row. Covered by a seeded chain id in the HTTP scenario. |
| CodeRabbit, `eth_sender_dal.rs:493`, rollback deletes a shared `eth_txs` id | Valid and reachable on the EN through the upstream reverter | Fixed: `delete_eth_txs` excludes ids referenced by retained batches. Covered by `delete_eth_txs(1)` in the HTTP scenario. Thread resolved. |
| Codex, `batch_status_updater/mod.rs:212`, explicit rejection never reaches blocked health | Valid | Fixed: the rejection branch runs before the zero-hash return; the health message covers both causes. The regression now uses the real rejected shape. |
| CodeRabbit, `storage_btc_block.rs:135`, require a timestamp before the main-node sentinel | Not valid | Declined: the main-node conversions read `via_l1_batch_inscription_request.updated_at`, which is `NOT NULL`, next to `is_finalized`; `via_en_executed_at` is not on that path, and the transaction conversion has no timestamp field. |
| CodeRabbit, outside diff, `blocks_dal.rs:1353-1355`, filter legacy sentinels out of the executed cursor | Deferred | Not applied: on an external node with legacy links the rewound cursor would make the updater re-link batches that are already linked, `set_eth_tx_id` refuses that, and the node would restart in a loop. The same query gates pruning and revert protection. The legacy-link question belongs to the historical recovery decision below. |

## Author checklist

- [x] PR title follows Conventional Commits.
- [x] Tests, docs and candidate formatting were updated; outstanding unchanged-file lint failures are explicitly recorded.
- [x] Ran `just via-check` and recorded results; no mapped sibling-pair path requires `via-check-strict`.

## Live infrastructure and deployment impact

No live infrastructure actions were run for this candidate's implementation, validation or publication. The only database writes were to isolated disposable local synthetic databases. Source changes can affect future main-node and external-node images; rollout to any live environment remains a separately approved deployment and verification step. PR CI builds are not deployments. Neither publication nor merge authorizes live migrations, restarts, rollback or historical writes.

### Migration and rollback

Apply the additive nullable `l1_batches.via_en_executed_at` migration **before** upgraded readers or writers use that database. Upgrade the main-node verdict producer before expecting external nodes to synchronize new synthetic executions. With an older main node, an upgraded EN withholds unsupported execution while commit/proof synchronization continues.

Prefer binary rollback with the added column retained. Older binaries do not maintain the new acceptance/timestamp contract; executions processed during a rollback interval can require authoritative revalidation after re-upgrade. Before applying the down migration, stop every upgraded reader and writer that depends on the column. Dropping it destroys all stored per-batch execution times, including any later recovered values. The shared synthetic history row cannot reconstruct distinct batch timestamps. No live binary rollback, schema rollback or service stop is authorized by this PR.

### Historical recovery decision

No historical retained-batch range is authorized here, whether incident-only or all retained batches. Recovery requires a separate reviewed one-shot operation and explicit operator approval. It must establish source/EN chain and genesis identities, settlement network, a matching non-null batch root and lifecycle hashes, affirmative source acceptance and that batch's exact execution time. Missing, rejected, pruned or mismatching evidence stays unresolved.

A stable source view must be coordinated against reorgs; a second RPC read alone does not close a subsequent remote race. Recheck local chain/batch/root, associations, related mutable history and row identity/version inside the write transaction; update only a still-NULL batch-local timestamp. A zero-row conditional update is a skip/revalidation result, not permission to blindly retry a replaced batch. No shared-history rewrite, fabricated inscription or cursor rewind is permitted. Identity mismatch, replacement/race and idempotence scenarios must first be rehearsed with synthetic data. Passing this PR's checks or merging it does not resolve the historical incident or authorize that operation.


romanornr:
@coderabbitai — Review feedback from romanornr on two findings at [be6a0a4](https://github.com/vianetwork/via-core/commit/be6a0a4e506afa6d00b29cfd472ab76367647b93).

Please reassess these against the source and identify any concrete counterexample or missed path. This is a request for case-specific clarification, not to add a persistent learning or a blanket rule suppressing similar findings.

### Execution timestamp guard — [original comment](https://github.com/vianetwork/via-core/pull/387#discussion_r3950950088)

This appears to conflate the main-node BTC reader with the EN mirror reader:

- `core/lib/dal/src/via_blocks_web3_dal.rs:50-51,118-119` selects `bir.is_finalized` and `bir.updated_at AS executed_at` from the same `via_l1_batch_inscription_request` row.
- `core/lib/dal/migrations/20240906134623_add_via_btc_inscription_requests.up.sql:26-33` declares that row’s `updated_at` as `TIMESTAMP NOT NULL`.
- The timestamp is optional in the query result because of the LEFT JOIN. When the joined row is absent, its verdict is absent too. Under this query/schema contract, an affirmative verdict cannot accompany a missing timestamp.
- The nullable `l1_batches.via_en_executed_at` belongs to the separate EN mirror readers, which already withhold sentinel execution when that timestamp is missing. The flagged main-node conversions do not read that column.

The proposed guard therefore appears unnecessary on the flagged path. Is there a reachable production query or schema change that invalidates this reasoning?

### Named execution sentinel — [original comment](https://github.com/vianetwork/via-core/pull/387#discussion_r3951591922)

A named constant could be a useful readability refactor, but the current source does not show a mismatched value. Four Rust sites agree: the producer, updater guard, per-batch timestamp writer discriminator, and mirrored-verdict conversion. Five SQL literals also match that value.

Changing only the two cited Rust sites would not unify the whole persisted/wire contract. The existing `via_settlement_status_sync_to_all_detail_apis` regression explicitly checks the `0x11…11` marker and distinct timestamps for two batches sharing it; preserving those literal expectations is important even if a constant is introduced later.

The proposed disposition is to defer this optional refactor to keep the bug-fix PR small, rather than treat it as a current major correctness defect. If there is a failing sequence under the current producer/consumer behavior—not a hypothetical future protocol-value change—please identify it so the disposition can be revisited.

These conclusions are based on source inspection and a separate read-only review. No new validation run was performed for this feedback.

